PR #6, opened 2016-11-21
Critical Typo fix

eaterjolly:
An inadvert-able double-negative critically and notice-ably changed the meaning of the line. Adjusted to present a grammatically stable triple-negative.


cgyarvin:
Thanks!

PR #25, opened 2019-01-16
Should this read "Stars and planets"?

deelawn:


flowerornament:
@deelawn yes, it should!

PR #27, opened 2019-01-16
Convert smart quotes to dumb quotes

Fang-:
Ideally we add a check to CI that rejects any PR containing smart quotes. /shrug


flowerornament:
@Fang- it would be cool if we also hooked up the CI to an airhorn in the office

flowerornament (approved):
LGTM

PR #32, opened 2019-01-17
Update README.md

rmariani:


rmariani:
I am keeping the fakezod boot-command the same. I tried with just `urbit -F zod -A <path/to/arvo/repo> -s` and I ran into an error.

PR #39, opened 2019-01-23
fixes inconsistent language

rmariani:
We use both "effects" and "actions" to describe the same thing. This change keeps it consistent.


flowerornament (approved):
LGTM

PR #52, opened 2019-03-12
Change the order of the getting started and introduction in the sidebar

vvisigoth:
Addresses https://github.com/urbit/docs/issues/431

### DIFF
--- a/partials/sidebar.hoon
+++ b/partials/sidebar.hoon
@@ -188,89 +188,6 @@
   ==
 ^-  manx
 ;ol.list-reset.h-font.mt-6
-  :: section: Introduction
-  ;li.mb-4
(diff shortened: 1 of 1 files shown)


flowerornament (approved):
I'm not sure I would do this myself – don't you need to find out what something is before you decide to install it? Nevertheless, I'm not too attached to which order these things end up in. Don't have time to test the build, but looks good to me.

rmariani:
urbit.org/docs still redirects to introduction.udon, so that still functions as a map of the docs. My plan is to have another page perform this same function, and having the "Introduction" section renamed to concepts.

flowerornament:
Nice! That sounds like an elegant solution. The introduction probably only needs to be a single page anyway, not an entire section.

PR #61, opened 2019-04-08
fixes broken links

rmariani:
links of the `[text](/child/grandchild)` variety don't work in urbit.org/docs/. It treats that as `urbit.org/child/grandchild` instead of `urbit.org/docs/child/grandchild`


flowerornament:
This checks in a /dist folder with a compiled copy of the website, which we don't want here

rmariani:
Fixed

flowerornament (approved):
LGTM

flowerornament:
Don't merge though please – I'll do it

flowerornament (changes requested):
Don't want to check this dist folder into the git history – please review https://github.com/urbit/urbit.org/pull/64 instead

PR #62, opened 2019-04-10
Simplified Chinese Primer

liweiz:
Not sure if the file name is ok, though.


matildepark:
@liweiz I'm really, really sorry that we've let this PR languish here for so long. We're about to relaunch urbit.org next week with a lot of new content, and so I'll have to close this, since the primer is being deprecated in the process.

That said, we really do appreciate this work and we'd love to have you do more translation — if you reach out to support@urbit.org we'd be glad to support your work more directly.

PR #82, opened 2019-05-14
Update the-100-year-computer.md

rmariani:
I made some minor changes. But I am worried that this is too terse to convey what we actually mean.

"Imagine a computer that was extremely valuable to you, that you’d pay thousands of dollars to recover, but you used *less* than the one you do today. "

This is confusing as hell. Are we saying that you would need to pay thousands of dollars to recover it? That that might happen? Why do you use it less than you use it today? Because it's just a better machine that needs less use? Are we saying that it's so valuable that you'd pay so much even for a device you rarely use?

So, general thought: we are trusting the reader to pick up on our hint too much.

### DIFF
--- a/content/posts/the-100-year-computer.md
+++ b/content/posts/the-100-year-computer.md
@@ -11,7 +11,7 @@ posts = ["Essays"]
 
 Around the Tlon office, we’re always looking for more concise ways to describe what we do. We kicked around ‘personal server’ for a long time. It’s pretty close. ‘Personal cloud computer’ did okay for a while, but it’s a little confusing. 
 
-But another way to think about Urbit is as a 100-year computer. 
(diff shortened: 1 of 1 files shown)


content/posts/the-100-year-computer.md
```diff
@@ -11,9 +11,9 @@ posts = ["Essays"]
 
 Around the Tlon office, we’re always looking for more concise ways to describe what we do. We kicked around ‘personal server’ for a long time. It’s pretty close. ‘Personal cloud computer’ did okay for a while, but it’s a little confusing. 
 
-One way to think about Urbit is as a 100 year computer. 
+But another way to think about Urbit is as a 100-year computer.
```
flowerornament: Prefer the "one" here

flowerornament (commented):
Merging this

Issue #67, opened 2019-05-08
Font inconsistencies 


g-a-v-i-n:
I'm noticing quite a few type inconsistencies in this new version.

- We don't use work sans anymore.
- The docs don't use the correct weights for headers. The type scale can be found in our Indigo file on Figma.
- Since the Primer has switched to Inter, the encircled number discretionary ligatures no longer convert.

flowerornament:
Thanks @gavinpatkinson. Assigning this to @friederjanmoerman and cc'ing @alexmatz.

- Where are you seeing work sans?
- I'll give @friederjanmoerman access to the type scales – we didn't make an attempt to match these in this revision, just a naive switch to InterUI
- How do you suggest we solve this? It looks like InterUI doesn't have the appropriate glyphs: https://rsms.me/inter/glyphs/ cc @jyng

g-a-v-i-n:
Ok, seems like the stylesheet was cached, no more Work Sans on my end.

Thanks for sharing the Figma with him.

I would just use numbering like 1. 2. etc

ohAitch:
(should there be a `contenthash` cache-busting filename for the CSS, or does the stylesheet change infrequently enough that it doesn't matter?)

flowerornament:
@ohAitch I'd love that, personally cc @friederjanmoerman

hardcodepunk:
Changing this to my suggestion would already style a lot to the right typography without me spending any time on it: https://github.com/urbit/indigo/issues/22

@gavinpatkinson 
Is it me or is it hard to determine the right typography parameters from this Figma file? I only see the font-sizes (also, two different numbers?). Is there a way to determine line-height, font-weight, margin and so on?

jyng:
@friederjanmoerman I can help you out here. If you look in the right corner of the Figma file, there should be a "Code" tab and that should give your CSS equivalents( font-weight, font-size, and line-height) on the item you selected.

hardcodepunk:
I had to make an account and log in to have the Figma functionality at my disposal.

The zola-port branch has been updated. Now the typography should follow the Indigo guidelines.
I also wrote a lil' vanilla JS script to style the input validation functionality. Now the newsletter `input` looks like the example in the Figma sheet:

![Screenshot 2019-05-13 at 00 54 12](https://user-images.githubusercontent.com/9070957/57588892-e2bb4000-751b-11e9-8ff7-62f10d20a0b8.png)

[...]

PR #87, opened 2019-05-17
RSS, post headers

flowerornament:


flowerornament:
Part of this PR addresses https://github.com/urbit/urbit.org/issues/67.

- My sense is that it's an extreme change that doesn't look great as implemented and needs some careful thought.
- I suspect @gavinpatkinson's intention was to have us use the Indigo type-scale just on posts and docs, but leave the homepage and menu as-is. This PR implements the type scale throughout the site, with the exception of page navigation (docs/post sidebar).
- Even if it were limited to posts and docs, I suspect the site would then feel unbalanced.

Please check out the preview here: https://deploy-preview-87--urbit-org.netlify.com/

flowerornament:
We've decided not to merge the typography updates. I'll pluck that commit out of this PR and put it on another branch.

flowerornament:
@friederjanmoerman branch with Indigo typography is now here https://github.com/urbit/urbit.org/tree/indigo-typography

PR #92, opened 2019-05-18
~rovnys-ricfer instead of ~rovnys-ricfur

yosoyubik:
Because ~rovnys-ricfur is a valid `@p`, @belisarius222 might want his contributions to be associated with his real identity


flowerornament:
thank you 🙏

Issue #83, opened 2019-05-15
User reported that the mailing list signup form doesn't work on Mobile Safari (iPad)


flowerornament:
A user reported that the mailing list signup form didn't work for them on Mobile Safari (iPad). Confirmed that the form works for me in Chrome and there have been a handful of signups over the past few days.

alexmatz:
can you please fix this one first @friederjanmoerman?

alexmatz:
looks like there's two issues here. 1. mobile safari/ipad is broken. 2. we just do not have any visual confirmation for users after they submit their email (i'll make another issue for this)

hardcodepunk:
_SSL certificate isn't recognized as 'safe'_ need to be fixed first #109

galenwp:
My comments from #109 are more relevant here. This is likely simply because we don't give user feedback on clicking subscribe, which is quite confusing. This is easy to test on desktop. Not sure what's up with iPhone 5 not understanding our cert.

friederjanmoerman:
@msutherl Can you mail me the original mailchimp subscribe code snippet from your mailchimp account, I do not have access to it? I'm suspecting something went wrong styling it. I want to debug the styled form because before it was working fine.

PR #110, opened 2019-05-26
Newsletter subscribe form bug fix

hardcodepunk:
Bug fix for #83 

The code snippet I was using before came from the old urbit website and is quite different. When I tested it, it was working. I don't know what changed in the meantime.
I replaced bits of code of a mailchimp embed code snippet I've used myself recently in a project and now it seems to work again.


galenwp:
Thanks for taking care of this so quickly

PR #120, opened 2019-06-13
Minor improvements to the grants post

Fang-:
Does what is says in the commit messages.

For the contributions, I think it's better to link to the lists of their PRs, since that's more permanent, and more interesting/useful to look at than just their profile/some graphs.


flowerornament (approved):
LGTM, thanks for these cc @alexmatz

alexmatz:
Thanks @Fang-!

PR #121, opened 2019-06-14
Divide all rems by 4 to fix Chrome rendering bug

### LINKED ISSUE #59: Minimum font-size set to 6px in Chrome 73
As of today, it appears that Chrome 73 is applying line-height: 8rem (rendered as 48px) to everything on the website. Safari doesn't appear to be doing this, nor Chrome Canary 75.

I wasn't able to reproduce on other machines, though we observed the same problem with Bridge.

- Line-height is too big in many places
- Menu icons are rendered larger than expected

The root problem appears to be that 8rem is being rendered to a larger pixel size than before.

Screenshot 2019-03-27 13 27 58

Screenshot 2019-03-27 13 30 07

Screenshot 2019-03-27 13 53 26

flowerornament:
Sets base font-size to default (16px) and divides all rems in CSS by 4 to fix https://github.com/urbit/urbit.org/issues/59.

### DIFF
--- a/sass/_main.scss
+++ b/sass/_main.scss
@@ -33,7 +33,7 @@ iframe {
 .wrapper {
   position: relative;
   margin: 0 auto;
-  padding: 0 (2rem / 4);
+  padding: 0 (4rem / 4);
   max-width: 960px;
 }
 
@@ -196,7 +196,7 @@ iframe {
   }
   &__nav-icon {
     position: absolute;
-    margin-left: (2rem / 4);
+    margin-left: (4rem / 4);
     width: (6rem / 4);
     height: (10rem / 4);
     left: 0;
--- a/sass/_main.scss
+++ b/sass/_main.scss
@@ -400,7 +400,7 @@ p, .p {
   font-size: (4rem / 4);
   font-weight: 400;
   line-height: (8rem / 4);
-  margin: (0 0 4rem / 4);
+  margin: 0 0 (4rem / 4);
 }
 
 pre, code {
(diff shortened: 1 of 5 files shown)


flowerornament:
@g-a-v-i-n @jyng @friederjanmoerman think it makes sense for me to merge this as a temporary fix? I haven't been able to look at the website with Chrome for over a month now, and who knows how many other people are suffering from the same bug.

g-a-v-i-n:
I don’t know the severity of the bug and it seems like an update will be avalible in about a week so I'm not sure its worth it.

flowerornament:
@g-a-v-i-n 

- When you say "an update," do you mean a Chrome update, or an urbit.org styling update?
- This seems to be the case, but in Indigo did we move away from the 4px root font-size?

Per severity, the website looks as below, with an extra .5em added to everything measured in rems. For me this bug shows up in Chrome and Chrome Canary. We know some other people are experiencing it because we submitted an issue here and others chimed in: https://bugs.chromium.org/p/chromium/issues/detail?id=949471

Screenshot 2019-06-14 13 0 34
Screenshot 2019-06-14 13 0 42

Per "worth it," after sleeping on it, I'm not sure there's any downside to merging this. Easy to revert later.

ohAitch:
Assuming the patch doesn't break things on like firefox

flowerornament:
@ohAitch tested and it doesn't – all the rem division happens in the scss transpiler, so there's nothing that would cause cross-browser compatibility issues happening here

g-a-v-i-n:
I'm talking about an urbit.org styling update where we don't use 4px root font size. Did you verify this works in Chrome and resolves the bug?

flowerornament:
@g-a-v-i-n great to hear! Tested on all browsers.

g-a-v-i-n:
Does it resolve the bug?

flowerornament:
@g-a-v-i-n it does

g-a-v-i-n:
ok cool LGTM

Issue #99, opened 2019-05-23
HTML title never displays any information except "Urbit"


rmariani:
"Urbit" is the only text that will ever display on the HTML title tab, no matter where you are on the website. It should display the page's title. 

<img width="209" alt="Screen Shot 2019-05-22 at 5 51 05 PM" src="https://user-images.githubusercontent.com/10265578/58218116-a70e3c00-7cba-11e9-8bed-62cde123b208.png">

matildepark:
@msutherl This can be closed now, too.

PR #129, opened 2019-06-26
Adding page titles to all templates (#99)

matildepark:
Addresses #99 by adding page titles to all templates — subbing in "section" and "page" variables where necessary in child templates and adding a block in the parent template. It appears to function correctly in my own testing.

Additionally the format is, for example, `Docs - Urbit`, to be as readable as possible. It can be changed if there's a preference for the styling of the title breadcrumb, of course.


flowerornament:
You rock – thank you.

PR #131, opened 2019-06-27
Homepage banner, embedded airtable and other details

francescanarea:


jyng (changes requested):
I think the styling here is good, but lets try to use the CSS from `indigo.scss` as much as possible. This makes it easy to track and maintain CSS from one source of truth. (If you need help getting acquainted with indigo, I can give you an overview)

I also noticed that the text inside `partials/top-alert` has `color:white`, but as you scroll the text underneath overlaps it. Let's try to fix that.

francescanarea:
yeah that'd be great.  come over whenever you're free

templates/partials/top-alert.html
```diff
@@ -1,12 +1,10 @@
...
+	<p class="white t-center mv0 ph8">
                 Interested in developing on Urbit? Sign up for Hoon School! 
-		<a href="/hoonschool" style="color:white">Learn more</a>
+		<a href="/hoonschool" class="white">Learn more</a>
 	</p>
 </div>
-
-<div style="padding-bottom:70px"></div>
-
+<div class="pv14"></div>
```
jyng: This is a good design change in that there needs more spacing between the top alert and the content, but its better practice to add margin to the top sibling (line 3). I also think `pv14` is a bit large and maybe something like `.mb4` would be better.

jyng (changes requested):

templates/partials/top-alert.html
francescanarea: Agreed about the spacing.  I did try the margin on the banner div but the issue is that it must have position:fixed so the margin isn't applied to the top of the body.  

I've done a bit of searching and can't seem to find a better solution.  Let me know if you might have any other thoughts

jyng: I made some small design changes here:
I think fixed actually makes it quite bothersome/distracting for the user if they're trying to scroll and navigate to a different page. So I removed it.
I fixed the spacing here to be a bit less so the bar is more subtle.

jyng (approved):
LGTM. I'm going to merge this.

flowerornament:
I made some significant changes to the design – smaller text, aligned with content.

PR #134, opened 2019-07-02
Application form

flowerornament:


flowerornament:
@francescanarea made a lot of changes here

- Re-implemented using `page.html` template
- Added headers
- Aligned top-bar text with content, but also commented it out for now

PR #135, opened 2019-07-02
Night mode site theme

matildepark:
Using the [prefers-color-scheme](https://developer.mozilla.org/en-US/docs/Web/CSS/@media/prefers-color-scheme) media query, adds SCSS to check if the browser or OS prefers night mode. If so, it replaces the site with a white-on-black theme. See pictured:

Screen Shot 2019-07-02 at 2 36 39 AM

How pleasant!

## The good

- Prefers-color-scheme is supported in Chrome, Firefox, and Safari (so most people). The others will just get the site as usual. Progressive enhancement, etc.

## The bad

- This implementation really isn't scalable if the site is going to change in a drastic way in future. It's simple, hardcoded and repeats values from the base theme in new locations (violating DRY), and even worse, overwrites classes like `bg-white` and `bg-black`. This would get confusing quickly. 

A better implementation would be having both versions of the site reassigning $background and $text variables, but it involves dancing with indigo; alternatively, outside of SCSS, adding and replacing our bg classes using JS (which is totally possible — just seems unwieldy in comparison, less seamless).

Because of the 'bad' section I understand if this PR should just be closed — and, after all, it was a quick experiment — but in any case the deploy preview should let you see the potential upsides, too. Let me know what you think.

### DIFF
--- a/sass/_night.scss
+++ b/sass/_night.scss
@@ -1,0 +1,16 @@
+@media screen and (prefers-color-scheme: dark) {
+    body.bg-white {
+        background: #000000;
+        color: #fff;
+        a {
+            color: #fff;
+        }
+    }
+    .bg-black {
+        background: #fff;
+        color: #000000;
+    }
+    .header__nav-icon:before, .header__nav-icon:after {
+        background: #fff;
+    }
+}
--- a/sass/styles.scss
+++ b/sass/styles.scss
@@ -1,3 +1,4 @@
 // custom
 @import "main";
 @import "responsive";
+@import "night"


matildepark:
An obvious question is how this would possibly tangle with the search-bar feature currently incoming, as well as other future releases. Hmm.

flowerornament:
Excellent work! Interesting questions. I might prefer this be implemented as a toggle, but then we need to deploy cookies. What do you think @galenwp?

matildepark:
It's why I really like the media query, actually — it sidesteps JS implementation, it goes off a system setting, it's automatic, it's very *native* in that sense. It's lightweight (only 17 lines to execute this) and adds very little heft to the site while providing a lot of utility. In any case, I'm open minded and curious to hear broad opinion!

flowerornament:
@matildepark I can definitely agree on all those points, but I fear some people prefer a light background even when their browser is in dark mode. This is an 'opinionated' decision that I think we should make carefully – I'll review next week.

matildepark:
I agree, it's definitely opinionated, so I feel you. Next week I should be around in person if it helps any.

edit: Actually, if it's a native flag we're parsing it seems less opinionated to follow their UI setting. Hmm. There's something to be said for our consistent identity though, which I think lends this to waiting for Galen's take.

matildepark:
For a bit of background in what other places are doing regarding this it looks like [Twitter](https://twitter.com/) is doing this — when my macOS is in dark mode, so is Twitter.

galenwp:
I've always liked the `cmd+option+i` photoshop-style command to invert the site colors. 

I think this approach is cute. Not super opinionated about setting cookies — but I could see some of our users feeling strongly about it.

matildepark:
Hmm, okay. Either way, this needs to be brought up to speed with the search release so I may need to open a fresh PR in any case (I deleted the source repo -_-). And since urbit.org might be seeing future design changes, I leave it to the @msutherl veto; we can either close this and we can be done with it or close this and I'll open another to update it.

PR #137, opened 2019-07-03
Disabling URL encoding of permalinks in RSS feed

matildepark:
The Zola base RSS template encodes permalinks. However, I can't see other RSS feeds doing this. Check for example [Stack Overflow](https://stackoverflow.com/feeds/tag/url)'s in comparison to [ours](https://urbit.org/rss.xml) right now.

This PR creates a base template that's the same but removes the encoding.

Before:

image

After:

image

Except, you know, it's not going to be localhost.

### DIFF
--- a/templates/rss.xml
+++ b/templates/rss.xml
@@ -2,7 +2,7 @@
 <rss xmlns:atom="http://www.w3.org/2005/Atom" version="2.0">
     <channel>
         <title>{{ config.title }}</title>
-        <link>{{ config.base_url | urlencode | safe }}</link>
+        <link>{{ config.base_url | safe }}</link>
         <description>{{ config.description }}</description>
         <generator>Zola</generator>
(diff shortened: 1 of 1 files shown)


flowerornament:
Pardon my ignorance, but would it make more sense to use the "description" from the post metadata for `<description>`, rather than the post content?

matildepark:
This is the exact same template we had before, just fixing the bug. If you want me to change where we pull that description from, I can do that while I'm at it.

Edit: Did a [quick look](https://www.wired.com/2010/02/rss_for_beginnners/).

image

That is, the description field is what RSS readers pull from when displaying the article. So we can have the summary and you click to go to the site, or we can just share the post and let people read on their clients, AFAICT.

flowerornament:
@matildepark K – whoever wants to own RSS can take care of this – thanks!

PR #139, opened 2019-07-09
Search bar release

francescanarea:
Fixed the implementation of search to account for partial word searches.  Simplified CSS for icons and misc search bar styling.  Tweaked search bar result styling.

### DIFF
--- a/sass/_main.scss
+++ b/sass/_main.scss
@@ -339,6 +339,14 @@ iframe {
  * styling
  */
 
+ // remove default x icon
+ input[type="search"]::-webkit-search-decoration,
+ input[type="search"]::-webkit-search-cancel-button,
+ input[type="search"]::-webkit-search-results-button,
+ input[type="search"]::-webkit-search-results-decoration {
+   display: none;
+ }
+
 textarea, input[type="text"], input[type="email"], input[type="number"], input[type="search"] {
   border: 1px solid #B1B2B3;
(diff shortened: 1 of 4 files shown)


flowerornament:
@francescanarea ok, this look pretty good, but a few things to fix before release

- The magnifying glass icon isn't vertically centered in its box 
- I'd replace the "X" icon with this glyph https://rsms.me/inter/glyphs/?g=multiply.case
- I think the light gray text is too light. I see the logic, but people with poor eyesight won't be able to read it at all
- The Search box as it appears on the page before you activate it is a little too tall. Simply removing the `line-height: 1.5rem` is a good fix.

Also, not sure why titles with Urbit dates render this way ...

Screenshot 2019-07-15 17 26 3

https://urbit.org/posts/2017-4-17-update/

francescanarea:
@msutherl 

see line 196 of scripts.js below for the date formatting issue:
  var lastPartTitle = pathArray[pathArray.length-2].replace(/-/g,' ');
the search result title is generated via the filename, rather than the title in the .md. 

the only way i can currently think of fixing this is by redo-writing how results are displayed by using a style similar to how we display posts in list.html 

Screen Shot 2019-07-16 at 4 56 30 PM

but perhaps you have a more elegant solution?

alexmatz:
@msutherl we ready to go on this? How can I help get this out?

PR #144, opened 2019-07-13
chmod -x on non-executable files

jfranklin9000:
janitorial


flowerornament:
@jfranklin9000 nice catch

PR #147, opened 2019-07-21
FAQ copyediting pass

matildepark:
Correcting some missing punctuation, fixing some use of capitals, and Zola was parsing the phrase `<some-planet>` like it was an HTML element, so I changed the formatting on that.


flowerornament:
@matildepark nice

PR #151, opened 2019-07-22
Prevent page refresh when using 'enter' key in search bar

### LINKED ISSUE #149: hitting enter closes the search bar
open the search bar, type something in, hit enter, search bar gets closed.

matildepark:
Fixes #149.

Unrelated -- all the search functions still have `console.log` in the code -- should this be removed? I can do that in another PR if necessary, it just makes the console a bit messy and if I had to guess was for development purposes.


flowerornament:
@matildepark good catch!

PR #153, opened 2019-07-25
Add hoon school banner to home page

francescanarea:
- delete templates/partials/top-alert.html
- direct html injection into index.html

PR #154, opened 2019-07-25
Re-add top-alert

flowerornament:


flowerornament:
@jyng I had made some style changes that I think improve the hoon school banner ("top-alert") in https://github.com/urbit/urbit.org/pull/153. Can you compare the deploy previews and let me know what you prefer?

cc @francescanarea

francescanarea:
@msutherl after I pushed this code I was confused about why it was commented it out so I thought that we didn't want it in a template.  I do think it's better to have it in its own file so that 1. it can taken out easily 2. the text can be changed and 3. it could be easily included in other pages.  sorry about that.

flowerornament:
No worries! I'm going to give this a day for comments and merge https://github.com/urbit/urbit.org/pull/154 by executive fiat if we don't hear anything.

francescanarea:
sounds good

PR #160, opened 2019-07-30
Fix bad redirect for /docs/pki

### LINKED ISSUE #927: urbit.org/docs/pki 404s
That link is mentioned on paper wallets and so should remain operational.

It redirects to urbit.org/docs/concepts/arvo-vs-azimuth/, but that just results in a 404. Probably want to point to urbit.org/docs/concepts/azimuth/ instead, but idk where the redirects live...

Fang-:
Fixes urbit/docs#582.


flowerornament (approved):
LGTM

PR #169, opened 2019-08-20
Fix janky History section

noqcks:
Before 
image

After
image

Also changing History # from 3 -> 4.

### DIFF
--- a/templates/primer.html
+++ b/templates/primer.html
@@ -813,10 +813,13 @@
             </div>
         </div>
         <div class="bg-gray">
-            <div id="azimuth-distribution" class="primer-container pt-40 mb-10">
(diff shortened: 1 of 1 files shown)


noqcks:
cc: @matildepark

matildepark:
1. Hello! Thanks for the cc.
2. @msutherl is the stakeholder for urbit.org so they'll have to merge it.
3. Good catch on ToC.
4. This looks solid and consistent with the rest of the primer formatting, so I'm personally happy with this PR.

flowerornament:
@noqcks thanks for this! Mind fixing `(5) Azimuth Distribution` as well while you're at it?

noqcks:
@msutherl fixed

PR #173, opened 2019-08-27
Replace Google Analytics with self-hosted Matomo analytics

Fang-:
This kills the Google. cc @vvisigoth


matildepark:
I admire your confidence, @Fang-, but Google is a hydra. That said I appreciate this step toward its containment.

Fang-:
Let's just say this is the first step in the process of curing ourselves of it!

PR #179, opened 2019-09-17
refactor docs, step 1

flowerornament:
This PR refactors docs according to the new IA:

- Deletes docs/getting-started
- Deletes docs/concepts
- Deletes docs/using
- Consolidates Hoon sections under docs/hoon
- Removes /docs/learn directory
- Removes /docs/reference directory
- Fixes all the broken relative links this created


flowerornament:
@rmariani heads up, here's the docs refactor

matildepark:
Seems fine; it's just that it will end up being undone once we merge the docs subtree again before launch. Between this and the [Hoon code blocks commit](https://github.com/urbit/urbit.org/commit/a00f2ea302bedabeb9fe32fd1a5c7c3f973a5a24), I feel like maybe we should backport the doc changes to docs before re-importing docs? 

Or else just manually inspect the changes to docs over this month and commit them ourselves, though we'd lose the commit history for docs for September by doing so. I don't think there's been that many commits to docs this month; so I don't think it would take that long to port.

matildepark (changes requested):
/docs still redirects to /getting-started, which 404s now.

flowerornament:
@matildepark indeed, we'll have to do it again – this is temporary

matildepark (approved):

flowerornament:
@matildepark hard for me to say what will be faster – I'll let you decide

PR #180, opened 2019-09-30
Next click tracking

flowerornament:
This should add event tracking to the mailing list subscribe button per https://matomo.org/docs/event-tracking/#tracking-events.

@matildepark there are other ways to implement this, i.e. add to every button with the right 'name' from `scripts.js` – let me know if you think there's a cleaner way to do this for now.


matildepark (approved):
Goodness, I thought this was a gigantic change from the email, but it was comparing it against master. Phew!

Seems fine, though I have no way to confirm functionality (no access to Analytics). We can talk about how many things we want to track; if it's sparse, inline is okay. If it's often enough, or on every version of a template page, we can create some template-specific hooks with basic loop iterations.

PR #184, opened 2019-10-09
Edited the intro copy and tighted the vertical spacing

vvisigoth:
I didn't do anything else on teh formatting tip, because I'm sort of confused by the code.

I contend that we shouldn't have what I consider to be the main copy in the figure captions, especially if we're going to have this little "drawing" intro. *BUT* I don't know hte most righteous way to do this, indigo-wise. Plz halp.


matildepark (approved):

vvisigoth:
cc @g-a-v-i-n @galenwp @jyng 

Screenshot 2019-10-09 10 50 15

PR #203, opened 2019-10-14
Copy edit for clarity

eriknewton:
@galenwp - I know this is lawyerly editing, yada yada. Anyway...


galenwp:
I don't love it, but maybe it's good

galenwp:
(Fine by me to merge this — not sure what's causing the checks to fail)

matildepark:
It’s because the checks were done only once, when the whole site was unable to build, midday, IIRC. Should be absolutely fine.

PR #204, opened 2019-10-14
Chore/polish

g-a-v-i-n:
**Footer**
- adjust footer width on medium screens
- make footer link text smaller to avoid overflow
- add consistent vertical margin between footer heading and footer link lists
- remove arrow for external links. They overflow a lot which makes the list hard to read. Can revisit this later.

**Video**
- Add `playsinline` to video tag - seeing if this enables autoplay on mobile. There are bandwidth concerns too, so I might remove this later.
- Add `poster` to video tag


templates/index.html
```diff
@@ -1,8 +1,7 @@
 {% extends "base.html" %}
 <!--External links are differentiated on home page.-->
 {% block templatestyles %}
-<!-- a[href*="//"]:not([href*="{{config.base_url | safe}}"])::after {
-content: " ↗"; -->
+
 }
```
matildepark: This `}` needs to be removed, will show on the site. Remove the comment on line 2 as well just to be tidy.

g-a-v-i-n: ah got it, good catch

matildepark (approved):

PR #205, opened 2019-10-14
Clarity changes

eriknewton:
@galenwp the language starting at line 75 was problematic for reasons we can discuss in person if you like.


galenwp:
Fine by me — not sure what got mangled here.

PR #206, opened 2019-10-14
add class wk-appearance-none to email signup

g-a-v-i-n:
- remove inner shadow in input box in safari


matildepark (approved):
Dope, but it needs to pull docs in so there's no more conflict.

PR #209, opened 2019-10-14
Language change

eriknewton:
@galenwp - this second sentence doesn't read as a continuation of the one on top of the video, so it needs to stand alone.


galenwp:
This is great

PR #211, opened 2019-10-14
search icon size and spacing, searchbar close size

g-a-v-i-n:
- move search icon to left justification 
- enlarge icon
- reduce margin between icon and 'search'
- enlarge close and search icons in searchbar


matildepark:
There’s a separate instance of the icon in templates/partials/navigation-docs.html.

PR #212, opened 2019-10-15
border css simplification/unification

urcades:
I basically removed some css classes since they implicitly already reflect black borders based on everything else being black already. 

The other big change here was the re-ordering of the statement erik pointed out as making more sense located at the bottom of the page.

I just realized all of the other heavy work was just asset swapping in GCP, so this pull request isn't actually that considerable, lol


urcades:
Going to fold one more commit into this PR: the sidebar nav, I just need to find the commit where I removed it.

g-a-v-i-n:
LGTM

g-a-v-i-n (approved):

PR #241, opened 2019-10-17
re-added page on booting a ship

drbeefsupreme:
as far as I can tell we lost this page in the shuffle. I don't actually know if the link to `/docs/concepts/azimuth.md` works or not, since its under the `/docs/` repo instead of `urbit.org` and I don't actually know how to test this, so I need someone who knows how to check this to do so. It looks right by comparing with how it is used in `using/develop.md`, though.

### DIFF
--- a/content/using/booting-a-ship.md
+++ b/content/using/booting-a-ship.md
@@ -1,0 +1,108 @@
++++
+title = "Booting a Ship"
+weight = 5
+template = "doc.html"
++++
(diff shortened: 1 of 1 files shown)


matildepark:
Basic Operations exists under /using/operations; but Booting Your Ship is missing ... It should go at the bottom of /using/install, but without this language (we recently rebranded a lot of this).

rmariani:
Yeah, [Install + Setup](https://urbit.org/using/install/) should include both install and booting. Can you make a new PR that edits that doc appropriately with this information, @syzygyzer ?

drbeefsupreme:
I will get to this later today

PR #245, opened 2019-10-17
instruction on how to boot your ship

drbeefsupreme:
Added instructions on how to boot your ship to the install+setup page. this is taken from the old `booting-a-ship.md`. I've changed the language a bit as well, in particular replacing point with Urbit ID.


content/using/install.md
```diff
@@ -74,3 +74,101 @@ If you're running Urbit in the cloud on a small instance, you may need to additi
...
+
+```
+./urbit -w sampel-palnet -G rAnDoMkEy
+```
+
+Either command will create a directory called `sampel-palnet/` and begin building your ship. It may take a few minutes.
+
+When your ship is finished booting, you will see the `~sampel-palnet:dojo>` prompt. At that point, you should permanently erase your keyfile from your machine.
+
+## Basic Operations
```
matildepark: Basically everything here below, IMO, should be removed, since it's duplicating content in Operations. I think since Docs owns this content, it's @rmariani's call to merge this stuff into the site.

rmariani: I think we should keep the Dojo part just so people get a taste for how to operate things. Everything below that should be cut from this doc. Another time we can talk about making an "Admin" doc that covers such material, @syzygyzer

content/using/install.md
```diff
@@ -74,3 +74,101 @@ If you're running Urbit in the cloud on a small instance, you may need to additi
...
+
+Welcome to your ship! Here's a few things you should do to become oriented. For a more complete overview of basic ship operations, see [Using Your Ship](@/using/operations/using-your-ship.md).
+
+### The Dojo
```
rmariani: Just make this a `##`-level header and replace `## Basic Operations` with it.

content/using/install.md
```diff
@@ -74,3 +74,101 @@ If you're running Urbit in the cloud on a small instance, you may need to additi
...
+Should produce:
+
+```
+> (add 2 2)
+4
+```
+
+Good.
+
+### Mounting
```
rmariani: Remove this line and everything below it.

content/using/install.md
drbeefsupreme: Yeah I considered that, but that page is pretty long and would be a tad overwhelming to someone new. Going through this short set of instructions should reassure a new Urbiteer that they installed everything correctly, and then they can go to the linked Operations page to learn more.

rmariani:
LGTM

PR #249, opened 2019-10-21
Simplify essay and UU grid

### LINKED ISSUE #247: Docs mobile template doesn’t use full page
Likely just changing the column classes on the two or three doc templates. 

4C5E7623-36DC-4E63-AE1C-029E521FA29C

matildepark:
This PR essentially brings these templates closer to my original implementation. When @urcades and I were adding image components that would be *full page* or *oversized compared to content* I had to make the grid a bit more complex to handle that possibility, adding a subgrid and some weird CSS selectors. Sorry!

Since we aren't using those components, this makes the CSS and templates much simpler. I also brought the docs templates to use the full grid on mobile in addition. I believe this closes #247.

### DIFF
--- a/sass/_main.scss
+++ b/sass/_main.scss
@@ -189,10 +189,6 @@ body.essay {
         display: block;
         // margin: 0 auto;
     }
-    article > *:not(figure):not(picture) {
-        grid-area: auto / 4 / auto / 10;
-        @extend .measure-wide;
-    }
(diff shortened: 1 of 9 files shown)


g-a-v-i-n (commented):
LGTM

g-a-v-i-n (approved):

PR #250, opened 2019-10-21
Feature: Twitter cards

matildepark:
I half-did this on Friday, but it required some tedious stuff, so it needed completion. I will do this properly now, in PR form. Gomen.

Uses OpenGraph when [the fallback is available](https://developer.twitter.com/en/docs/tweets/optimize-with-cards/overview/markup). Adds the appropriate meta tags for all templates.

### DIFF
--- a/templates/base.html
+++ b/templates/base.html
@@ -5,12 +5,13 @@
     <meta charset="UTF-8">
     <meta name="apple-mobile-web-app-capable" content="yes">
     <meta name="viewport" content="width=device-width">
-        <!-- TWITTER -->
-        <meta name="twitter:card" content="summary_large_image">
-        <meta name="twitter:site" content="@urbit">
-        <meta name="twitter:creator" content="@urbit">
-        <!-- END TWITTER -->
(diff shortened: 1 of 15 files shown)


templates/base.html
```diff
@@ -5,12 +5,13 @@
     <meta charset="UTF-8">
     <meta name="apple-mobile-web-app-capable" content="yes">
     <meta name="viewport" content="width=device-width">
-        <!-- TWITTER -->
-        <meta name="twitter:card" content="summary_large_image">
-        <meta name="twitter:site" content="@urbit">
-        <meta name="twitter:creator" content="@urbit">
-        <!-- END TWITTER -->
-    <meta name="description" content="{% block description %}{% endblock description %}">
+    <meta name="twitter:card" content="summary_large_image">
```
g-a-v-i-n, line 8: What image is used here?

g-a-v-i-n (approved):

templates/base.html
matildepark: image

PR #252, opened 2019-10-22
Feature: Docs template tweaks, adds automated "next up" pagination

matildepark:
This PR adds "next up" pagination for all non-chapter, non-section pages.

It also amends the docs template — it was too narrow on iPad, and wasn't overall very responsive, so it now defaults to single-column on iPad resolution; uses more of the grid by default; and adds `measure-wide` to match the rest of the site.

To be deployed alongside [docs/672](https://github.com/urbit/docs/pull/672).

image

### DIFF
--- a/templates/doc.html
+++ b/templates/doc.html
@@ -12,7 +12,7 @@
 <!-- header -->
 <!-- content -->
 {% include "partials/navigation-docs.html" %}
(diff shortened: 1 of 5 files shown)


g-a-v-i-n (approved):
Sorry for the protracted review, LGTM.

PR #259, opened 2019-10-28
Add gray background to inline code

### LINKED ISSUE #253: Can we restore the gray background for inline code?
The code font and the script font look very similar and the different background would help see these better.  We used to have it, and we still do for block code.

matildepark:
Adds `.bg-gray5` to inline code blocks, while not adding them to dates and times and non-code monospace use. This matches the background color already in use on bigger code blocks. 

This PR also includes a docs update while we're at it.

Closes #253.

### DIFF
--- a/sass/_main.scss
+++ b/sass/_main.scss
@@ -408,6 +408,10 @@ code, .code, .mono {
     font-family: "Source Code Pro", monospace;
 }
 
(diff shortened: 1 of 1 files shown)


g-a-v-i-n (approved):
LGTM

Issue #246, opened 2019-10-17
cannot search for Ford runes


drbeefsupreme:
Ford runes begin with `/`, but typing `/` into the Search bar closes out of it.

matildepark:
We can change the hotkey easily enough, but [Ford runes aren't in runes.js anyway](https://github.com/urbit/urbit.org/blob/master/static/runes.js). Needs to be added, if you want them — since they're not listed as runes on the runes pages, only on Ford's page, they weren't processed.

drbeefsupreme:
I'll discuss what we should do about this with the docs team on Monday, since they ought to be separate from Hoon runes somehow, but they also ought to be searchable.

matildepark:
You could just add them to runes.js without moving the docs around? The hotkey thing is my part — We just change the hotkey event listener in scripts.js.

PR #261, opened 2019-10-29
Adds Ford runes to runes.js

drbeefsupreme:
This addresses #246 , but also needs @matildepark to adapt the search bar, and for the ford-headers branch to be merged in the docs repo for the links to work.

### DIFF
--- a/static/runes.js
+++ b/static/runes.js
@@ -853,4 +853,123 @@ var runes = [
       "link": "/docs/reference/hoon-expressions/rune/terminators/#tistis",
       "desc": "The <code>==</code> rune is used to indicate the end of a 'jogging' or 'running' series of Hoon expressions."
     }
+    {
+      "name": "faslus"
+      "symbol": "/+"
+      "usage": "Ford"
+      "link":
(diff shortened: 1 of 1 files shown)


matildepark (approved):
LGTM. What should we change the keyboard shortcut to? Suggestions? @urcades ? @g-a-v-i-n ? `⌘/` is potentially nice, but is the Windows key / Meta key on non-Mac platforms, and less intuitive. We also don't mark a shortcut literally anywhere on the site. We could just strip the shortcut, but power users will appreciate one.

Fang-:
imo only pressing escape should close the search bar (and return should go to the selected/first result).

matildepark:
But what about _opening_ the search bar?

Fang-:
That can remain unchanged just fine, right? Maybe I'm mistaken, but I feel "x to open, escape to close" is pretty common.

matildepark:
So:

`/` to open
`esc` to close
`Enter` to hit first result

I think I can sneak this in today, then.

PR #262, opened 2019-10-29
search: revise keyboard shortcuts

### LINKED ISSUE #246: cannot search for Ford runes

matildepark:
- `/` opens search window
- `Esc` closes search window
- `Enter` goes directly to first search result's page

Closes #246.

### DIFF
--- a/static/scripts.js
+++ b/static/scripts.js
@@ -15,11 +15,13 @@ function initHotKeys() {
         document.body.classList.remove('has-active-search-window');
       }
 
-      // / to close toggle window
+      // / to open toggle window
       if (key === '/' || key === 'Period' || key === 58) {
-        event.preventDefault();
-        document.body.classList.toggle('has-active-search-window');
         if (bodyEl.classList.contains('has-active-search-window')) {
+          // if it's already open don't do anything!
+        } else {
+          event.preventDefault();
+          document.body.classList.add('has-active-search-window');
           searchInput.focus();
         }
       }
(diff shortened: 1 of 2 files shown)


static/scripts.js
```diff
@@ -30,6 +32,14 @@ function initHotKeys() {
           searchInput.focus();
         }
       }
+
+      if (bodyEl.classList.contains('has-active-search-window')) {
+        if (key === 'Enter' || key === 'Return' || key === 13) {
+          event.preventDefault();
+          let firstresult = document.querySelectorAll("div#full-results a")[0];
+          firstresult.click();
+        }
```
Fang-, line 41: Since you're touching this rn, how hard would it be to allow the user to select results using the up-down arrow keys? Just a matter of catching those and storing/modifying an "index" state variable?

If it's trivial, would be great to have that added. If not, probably out of scope here. (I might already be guilty of scope-creeping this PR.)

matildepark: *shakes fist at `~palfun`*

matildepark: Thought for a sec. I do think it's a touch out of scope at the moment, since we're dealing with multiple arrays of results, and four new key events plus now tracking state (across search sessions on the same page as well). Would have to rethink search data structure for a sec and then test. Like it's not even really that hard, it's just that this is just meant to be a chore PR.

Fang-: Still sounds like good news to me. Very fair! (:

g-a-v-i-n (approved):
LGTM

PR #265, opened 2019-10-30
add lh-copy to docs nav

matildepark:
The /docs navigation sidebar has some weird spacing because multi-line titles are spaced differently than each element's `margin-top`. Giving the nav list `.lh-copy` spaces things evenly.


g-a-v-i-n (approved):
LGTM

PR #266, opened 2019-10-31
add cta + bridge info

matildepark:
Adds a mailing list CTA to the bottom of the last Understanding Urbit post (previously had nowhere we wanted readers to go or do once finished the series).

Adds Bridge information and "what to do once you have an urbit ID" to Install + Setup above the binary instructions, since that information was missing. Bridge is a common outlink from the site, but it's presently buried a bit deep in the site, and it isn't in the footer, so people are looking for it somewhere. 

(We discussed putting it on the front page alongside "Docs" and "Install" but it feels more natural as a small part of "Install".)

### DIFF
--- a/content/understanding-urbit/project-history.md
+++ b/content/understanding-urbit/project-history.md
@@ -39,7 +39,7 @@ At the beginning, Urbit was just a few people with the right combination of imag
 
 ## Future updates
 
-Urbit is young, but developing quickly. If you've just finished our Understanding Urbit series and want to stay connected with the project, we recommend starting with our mailing list. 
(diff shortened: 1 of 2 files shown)


galenwp:
My hope was to add a sort of 'coda' — a final page here where we direct people to the mailing list, docs or other channels to follow us. I'll get to that eventually. For now I just added the GH and Twitter links, and I think this is good to go.

galenwp (approved):

matildepark:
Coda page would be pretty trivial to add — just another .md page in the directory with a weight higher than Project Status's and the template components would handle the rest.

galenwp:
Yep, I know — just gotta actually write the content. Or at least outline it.

PR #267, opened 2019-10-31
Feature: Documentation navigation auto-scrolls to latest open navigation drawer

matildepark:
This was something I PR'd into the last incarnation of the site, but the functionality's changed a bit. 

Links no longer have a `scrollHeight`, but the drawers do, so this just scrolls the latest open drawer into view on page load. Only really noticeable when using the Hoon tutorials or standard library (and really, only needed there, too, since most drawers are tucked out of the way, otherwise). Just some polish.

(I'm doing docs PRs because I'm using them and hitting annoyances, heh.)

### DIFF
--- a/static/scripts.js
+++ b/static/scripts.js
@@ -325,7 +325,16 @@ if (document.readyState === "complete" ||
   });
 }
 
+// Scroll to current document in nav list.
+let docsNavScroll = function() {
+  let docsDetails = document.querySelectorAll("nav ul details")  
+  for (let details in docsDetails) {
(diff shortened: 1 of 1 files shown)


matildepark:
(This in particular works *really well* with the rune search, since the list of rune types are automatically unfolded and opened for you. You can crack open urbit.org's search, type a rune, hit enter and have the docs split open to the exact page, section, and drawer. Slick.)

g-a-v-i-n (approved):
LGTM

PR #273, opened 2019-11-04
docs: center dropdown on mobile

matildepark:
Centers the dropdown for navigation on mobile. It's a gross, hard-coded way to do it, but it's really bugging me, and I didn't want to use JS to do it — pure CSS solutions work on everything but Safari, so iOS users would always see the weird left-aligned one.


g-a-v-i-n (approved):
I think this is a totally appropriate shim.

PR #276, opened 2019-11-06
20191106 Weekly(?) Update Post

urcades:
Adding a new post to the /updates page.

Would like some eyes on how links are displayed, copy, image selection, etc.

Related question: Do we have rules on how urls/links/copywriting is handled? I noticed urls were handed to me bare, but previous updates generally try and hide urls under 'normal' writing.

Please advise.

### DIFF
--- a/content/updates/2019-11-06-update.md
+++ b/content/updates/2019-11-06-update.md
@@ -1,0 +1,38 @@
++++
+title = "~2019.11.06 Update"
+date = 2019-11-06
+description = "Weekly Web Update"
+aliases = ["/posts/updates/2019-11-06-update", "/posts/2019-11-06-update/"]
+[extra]
+author = ""
+ship = ""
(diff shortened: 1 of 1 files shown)


crrncllns:
urls were in that format for a plain text email and i didn't change them - sorry

urcades:
@collinscorrina Gotcha! It's no problem, was curious about how we go about that styling in general.

galenwp:
This is definitely written in the style of an email update (which is plain text). I'm perfectly comfortable rephrasing all of this such that it actually uses [link](https://urbit.org)s

crrncllns:
@urcades should be consistent with previous updates to that page. not sure what @galenwp has in mind.

matildepark:
Following @galenwp's advice, amend the plaintext links to [links] and feel free to merge.

PR #286, opened 2019-11-15
Add posts' leading image as Twitter card image

matildepark:
Just warming up! We were previously showing the sig as our graphic on all pages, which is fine but not ideal. We have a lot of beautiful graphic content that makes for a captivating pull when sharing links to site content.

This PR adds the template components for specifying an image and (and the sig default) when generating the page's meta tags, and then adds the metadata on all applicable posts. (I also ended up cleaning up broken Markdown syntax as I found it.)

### DIFF
--- a/README.md
+++ b/README.md
@@ -51,6 +51,9 @@ author = "Matilde Park"
 <!-- Ship: The author's ship. Used on post directory pages. -->
 ship = "~haddef-sigwen"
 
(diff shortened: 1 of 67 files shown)


g-a-v-i-n (approved):
LGTM

matildepark:
😎

image

PR #288, opened 2019-11-15
add blog post: Why Hoon?

matildepark:


galenwp:
Still reading — but I'm not so sure about this image. What is it supposed to be?

content/blog/why-hoon.md
```diff
@@ -8,7 +8,7 @@ ship = "~rovnys-ricfer"
 image = "https://media.urbit.org/site/posts/essays/hoon.png"
 +++
 
-![](https://media.urbit.org/site/posts/essays/hoon.svg)
```
Fang-: Maybe there's a trick we can use to coerce this into desired width?

(Not telling you to find that out and fix it right now, but maybe it needs to become an issue? SVGs are great for many reasons, we should want to be able to use them.)

matildepark: The native svg was 480px. The png, exported at 2x, was 960px. The site uses a max-width: 100%, it’s just a small svg. Not a general svg bug.

urcades:
Hi @galenwp — was playing around with a few possibilities, but I wanted the image to represent "the encapsulation of Nock by Hoon" — Nock here being represented with 12 squares indicating its limited opcodes, and Hoon being the enveloping squircle.

I get that it's a reach to expect anyone to actually read into the image that much, so it was more meant to feel like a simple geometric piece. 

Happy to revise it conceptually or use a Landscape/something else to stand in for this image, I get that it's a bit of a stretch, and we still haven't collectively landed on a way to represent nock/hoon via imagery.

galenwp:
That all kinda makes sense — it just looks like an ios app tile at first. Perhaps this would be less true if the edges of the enclosing square weren't rounded, or if the internal checkered area was somehow different.

urcades:
Yeah, I can give it another pass in terms of the shapes being used — originally the squircle was a circle, which I can revert to:

Screen Shot 2019-11-15 at 4 50 06 PM

Some other tinkering is located here:
[figma](https://www.figma.com/file/ESjGzsC6XWMw5p4aZ7k6GUrW/Illustrations?node-id=46%3A2)

galenwp:
Pushed a few minor fixes. The text is great. I riffed on Ed's designs a little and I think we have something that will work.

matildepark:
Then the asset can be replaced on GCP and this can be merged once that’s confirmed.

jyng:
Screen Shot 2019-11-15 at 2 26 52 PM
Seems like a markdown error here

matildepark:
That isn’t an error, it was all inline and I made it into a footnote below that particular blurb.

matildepark:
ie. here's the gdoc draft's:

image

jyng:
Oh, Got it. I guess it would seem less a mistake if it was italicized?

galenwp:
I almost changed it to be a parenthetical.

joemfb:
IMO, it's an unnecessary detail. It's just the way you go about incrementally making them  "share structure whenever possible".

belisarius222:
Yeah, good point, let's just lose the footnote

—
~rovnys-ricfer
https://urbit.org

matildepark:
Alright. I’ll cut and push.

belisarius222:
One thing I'd consider changing is the word "constrained" in the last paragraph. It's not exactly right: I was trying to say something more like "we use far fewer primitives and each one is closer to the fundamentals of what it is, without extra options or metadata". I'm having trouble phrasing that in a way I like, though

—
~rovnys-ricfer
https://urbit.org

joemfb:
It's not exactly right, but I think the context makes it clear. I don't know how to put it better.

PR #289, opened 2019-11-16
Hosting a meetup tips

rmariani:
Adds a page on best practices for hosting a meetup

### DIFF
--- a/content/community/_index.md
+++ b/content/community/_index.md
@@ -1,0 +1,5 @@
++++
+title = "Community"
+sort_by = "weight"
+template = "list.html"
++++
--- a/content/community/governance.md
+++ b/content/community/governance.md
@@ -1,5 +1,8 @@
 +++
 title = "Governance"
+description = "On the governance of the project, including both the senate of galaxies and the core developers."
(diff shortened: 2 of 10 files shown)


matildepark (commented):
This functions, but I wonder if this needs to be its own page, really. Why not just the bottom half of /meetups? If we use the page_indiced template it’ll be really clean and navigable too.

rmariani:
I considered that, but the team and I think it's unwieldy. I am not against it as long as it's clean! How would that be done?

matildepark:
I gave it a shot and you're right, it looks unwieldy. The best solution I found in my local environment is to make a /community/ directory and put the following pages in it:

- Hoon School
- Meetups
- Hosting a Meetup
- Governance

This matches the community IA to the other headings in the footer and gives it all a home to refer to instead of these two dangling meetup pages on our root directory.

I have this working if you're cool with me pushing it onto this branch and merging it all in.

rmariani:
I'm fine with that. Thanks a lot!

matildepark (approved):

PR #300, opened 2019-11-20
pad->pam, vat->pat, net->fas

drbeefsupreme:
fixes some pronunciation changes. Am I missing any?

Some of these are right in /docs/, some are not, I will be making a PR shortly in the docs repo to get the stragglers

### DIFF
--- a/static/runes.js
+++ b/static/runes.js
@@ -280,10 +280,10 @@ var runes = [
       "desc": "<code>[%wtts p=spec q=wing]</code>: test pattern match."
     },
     {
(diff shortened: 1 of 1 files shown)


drbeefsupreme:
https://github.com/urbit/docs/pull/729

ohAitch:
Now for the hoon compiler…

jfranklin9000:
I have an old bus->buc patch for hoon.hoon (and a few other files).  (That was an easy one.)
I'll try to find it and see if it still works.

PR #306, opened 2019-11-27
remove left/right margin on post page lists

matildepark:
I was going a bit batty over how articles and post directories sat on the grid differently. Turns out some list margins were getting in the way. This fixes that. Makes the post pages look better on mobile, too.

(To check this PR, just go to a posts page and a blog post and compare.)

### DIFF
--- a/sass/_main.scss
+++ b/sass/_main.scss
@@ -257,10 +257,12 @@ ul.posts, nav ul, nav li {
     list-style-type: none;
 }
 
-ul.posts {
-    li {
-        margin: 1rem 0 1rem 0;
-    }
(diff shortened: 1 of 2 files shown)


urcades:
God I was noticing that too while working on the last post in particular — this change is excellent and needed and let's just do it now.

A small thing I noticed while checking this out:

https://www.dropbox.com/s/j05lrpkd6bbl0q0/Screen%20Recording%202019-11-26%20at%207.17.17%20PM.mov?dl=0

matildepark:
Yes, slight jump. I can peek.

matildepark:
OK. *Should* fix that?

PR #309, opened 2019-12-02
Adding rough copy from Figma

urcades:


templates/index.html
```diff
@@ -44,9 +44,10 @@
 <!-- <nav class="full-bleed full grid12"> -->
 
 <section class="full c2-7-md c3-7-lg mt6">
-  <p class="lh-copy mt1 mb4">
-    Urbit is a simple, durable computer that you own 100% and forever.
-  </p>
+  <p class="lh-copy mt1 mb4">Urbit is Urbit OS and Urbit ID.</p>
+  <p class="lh-copy mt1 mb4"><b>Urbit OS</b> is a computing environment you can run anywhere. It’s built on top of a tiny low-level virtual machine built from scratch and designed explicitly to make actually-personal computing accessible to everyone.</p> 
+  <p class="lh-copy mt1 mb4"><b>Urbit ID</b> is a PKI built to replace DNS and permit single-sign-on-like functionality for accessing one’s Urbit OS.</p>
```
joemfb, line 49: I think this summary of "Urbit ID" concedes far too much. While more-or-less correct, DNS and SSO are just a couple of many use-cases for our totally-generally pseudonymous identity system and associated PKI (which are two separate things). Happy to discuss elsewhere if need be.

galenwp: This is super rough — I'm about to start working in this branch to iterate on copy.

urcades: Yes — there's quite a bit of clarification that needs to occur here. This PR was materialized by a recent meeting in which a few of us (@galenwp , @eriknewton , @loganallenc , and @matildepark included) were striking out to heavily clarify the most interesting components of "Urbit, the project" and communicate these components in a mostly abbreviated manner.

I think the hardest part about all of this, to your point, is the process of surfacing the more "relevant to the public" aspects of Urbit at the cost of seemingly simplifying the system. Like any true fan of what Urbit is doing, I'd love for people to appreciate the project for the totality of what it's trying to accomplish and restructure, but as it's been recently noted, we may need to concede key aspects of the project's total impact ... at least on the front page.

I think @galenwp will be resurfacing over this present copy and likely transform it into something more "total", but we shall see.

matildepark:
I think with #329 this can be closed, yes? @urcades

urcades:
Yeah, this seems correct.

PR #314, opened 2019-12-06
Post 2019-12-06-update.md to Updates

urcades:
Just propping this up for preview/review purposes. Will merge when all work threads are settled.

### DIFF
--- a/content/updates/2019-12-06-update.md
+++ b/content/updates/2019-12-06-update.md
@@ -1,0 +1,45 @@
++++
+title = "~2019.12.06 Update"
+date = 2019-12-06
+description = "Weekly Web Update"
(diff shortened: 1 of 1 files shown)


matildepark:
Should be ~2019.12.6.

matildepark:
Also nb when we do updates with two double digit numbers (eg 12/16) it removes the date from the permalink, hiding it from the site under the slug “updates/update”. You end up needing to add slug metadata like this post has:

https://github.com/urbit/urbit.org/blob/master/content/updates/2018-12-18-update.md

When a post has a date with a single digit and no leading zero (most of them) it doesn’t remove the date from the permalink.

PR #315, opened 2019-12-10
add community grants

matildepark:
Embeds an application for community grants on its own page in the /community section. Adds some copy. Replaces "Hosting a Meetup" in the footer, since hosting a meetup is more or less related as a pair with "Meetups" generally.

Also renames the link to grants.urbit.org in the footer to "Developer Grants", though I know that's not strictly true. We may need a better name than a Community Grant for this page, in such case. Let me know what you think.

https://deploy-preview-315--urbit-org.netlify.com/community/community-grants/

### DIFF
--- a/content/community/community-grants.md
+++ b/content/community/community-grants.md
@@ -1,0 +1,14 @@
++++
+title = "Community Grants"
+description = "Select Urbit address space grants for your group or community."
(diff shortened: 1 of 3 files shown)


bioy-casares (approved):
lgtm

PR #317, opened 2019-12-10
Add glossary to documentation, navigation, and site search

### LINKED ISSUE #293: "Using" section too separate from docs?
(I think this goes here instead of the docs repo.)

Maybe I'm blind, but it seems like the "using" section of the docs isn't actually reachable from the sidebar as displayed on [`/docs`](https://urbit.org/docs/)? Perhaps related, the root page is at [`/using`](https://urbit.org/using/) instead of the more-expected `/docs/using`.

Did we split "using" from "docs" intentionally? The "user docs" vs "tech docs" split makes _some_ sense to me, but certainly they should still link to each other. (Especially consider the `/using` pages are linked much less prominently from the homepage than `/docs`.)

matildepark:
Closes #293. 

- Adds templates and custom navigation for the glossary within the Urbit documentation. Moves usage information to the top of the documentation sidebar once again.
- Glossary template generates its link list programmatically, which is then sorted alphabetically for each category.
- Renames runes.js to glossary.js and adds the glossary definitions to site-wide search.

image

image

image

## Deployment

This branch is pulling its copy of the docs from [its "glossary" branch](https://github.com/urbit/docs/tree/glossary). To make corrections to copy on the docs, edits must be performed there and pulled back into this preview copy of urbit.org with a submodule commit.

To merge this PR, the glossary branch should be merged into master in the docs repository.

After that, this branch will once again check out master and merge.

### DIFF
--- a/static/glossary.js
+++ b/static/glossary.js
@@ -1,4 +1,5 @@
-var runes = [
+var glossary = [
+  // runes
     {
       "name": "dot",
       "symbol": ".",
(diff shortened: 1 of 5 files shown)


matildepark:
Here's a [direct link to the deploy preview of the new Docs](https://deploy-preview-317--urbit-org.netlify.com/docs/).

matildepark:
@rmariani Offhand stuff I already have:
- Landscape is still discussed as if it is old Landscape. Modulo probably shouldn't be a term.
- Urbit Wallet Generator is discussed as if in progress. I thought that was released. I am unsure.
- A "breach" isn't defined.

rmariani:
Thanks so much, Matilde, this is great. And good points, I will address those.

PR #319, opened 2019-12-10
Point to PR for document contents

Fang-:
Also makes the Bridge link prettier.

This allows us to get a bit more contextual information across.


vvisigoth (approved):
LGTM

PR #325, opened 2019-12-15
essay template: author + address margin fix, slight color tweak

matildepark:
If a post had multiple authors, the margin on the left of `<address>` element would push it slightly askew from the date and author name.

This just places it on the author's name instead, so they're always lined up.

This PR also changes the ship's color to `gray0` to, like the date, more subtly segregate the metadata from the author name.


urcades (approved):
These are great updates – the color change in particular is pretty ⚡️

PR #333, opened 2019-12-23
docs: add hoon template

### LINKED ISSUE #269: feature request: dynamically generated lesson, exercise, etc. numbers
The Hoon school lessons are under constant revision, and as such, the lesson numbers frequently change and so any references to them must also be changed manually. Exercise numbers, written as "Exercise 1.7a" for the first exercise in Lesson 1.7, also depend on these lesson numbers. With the planned expansion on adding figures, additional exercises, etc., a general purpose reference system that generates lesson, exercise, figure, etc. numbers automatically according to their placement in the table of contents will cap off an unbounded source of busywork.

Besides titles for these objects, we should also be able to reference them from elsewhere in the text, like "See Figure 1.4.4" being a hyperlink to that figure.

matildepark:
Closes #269.

Adds a basic programmatically generated index page for the Hoon tutorials. Nb. It hard codes the chapters, and dumps the remainder lessons and appendices afterward. But at least there's no need to keep track of this by hand anymore.

image

### DIFF
--- a/templates/sections/docs/hoon.html
+++ b/templates/sections/docs/hoon.html
@@ -1,0 +1,52 @@
+{% extends "base.html" %}
+{% block title %} 
+  <meta name="og:title" content="{{ section.title }} - {{ config.title }}">
+  <title>{{ section.title }} - {{ config.title }}</title>
+{% endblock title %}
+
+{% block description %} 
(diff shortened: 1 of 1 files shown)


rmariani (approved):
LGTM, thanks Matilde

PR #334, opened 2020-01-06
adds Toronto meetup

rmariani:


matildepark:
*ON

(Also we have one in Toronto? What?)

matildepark:
Also I tend to alphabetise by "state" ...

rmariani:
Fixed capitalization and alphabetization. And yep! Would love to hear how it goes

PR #336, opened 2020-01-16
Add a couple of small improvements to the docs.

waldenraines:
* Make it clear that /path/to/urbit is the path to the executable
* Add instructions for getting chat-cli connected


rmariani:
Thanks for the improvements!

PR #337, opened 2020-01-21
essays/list.html: add images

matildepark:
Adds images to the blog post directory page.

Please review the design and specify if you'd like a default image, too. If an image currently used here is hella ugly, let me know and I'll just remove it from the post's metadata. Anything ugly here will be ugly on an OpenGraph card.

### DIFF
--- a/content/blog/landscape-a-portrait.md
+++ b/content/blog/landscape-a-portrait.md
@@ -6,7 +6,6 @@ aliases = ["/posts/essays/landscape-a-portrait/", "/posts/landscape-a-portrait"]
 [extra]
 author = "Matilde Park"
 ship = "~haddef-sigwen"
-image = "https://media.urbit.org/site/posts/essays/landscape-a-portrait-1.png"
 +++
 
 ![](https://media.urbit.org/site/posts/essays/landscape-a-portrait-1.png)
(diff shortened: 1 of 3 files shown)


urcades:
I like it, go go go go gog

urcades (approved):
Looks good to me — I'll submit a basic default image in a coming incremental update

matildepark:
image

Can I just remove the image on this one? Looks so weird.

urcades:
yeah, let's just get rid of it — most of these images were developed to be squares, so this sudden rectangular cropping is probably alien for most of them

matildepark:
Would you prefer it if I didn’t do the rectangular masking and instead just displayed the full square image? It just seems less “preview” then, was my intuition.

— 
~haddef-sigwen
https://urbit.org

urcades:
Heh, yeah ... it might be a lot to display fully. 

Had I known we'd been retooling the blog index to display images, I probably would have just made the entry images rectangular — I'm curious to check out a version of this PR that uses the square cropping — I'd wanna know if it's too overwhelming for a simple index.

matildepark:
image

S'kinda big.

matildepark:
Assuming that we're not going with the big square images — we need to update the directory layout, since it's too vertically spaced without the images, so I'll merge this.

PR #339, opened 2020-01-23
even glossary padding, use indigo class

matildepark:
Uneven highlighting, top had less padding than bottom. Not a good implementation on my part. Cleaner now in all respects.

Before:

image

After:

Screen Shot 2020-01-23 at 5 03 41 PM


matildepark:
@urcades any objections?

urcades (approved):
It's looking beautiful, good improvement, let's merge it.

PR #341, opened 2020-01-28
Using Bridge: some touch-ups

Fang-:
Tried finding this in the docs repo first. Confusing.

This changes the headers for various ways of running Bridge to be more accurate. There is no "offline" version of modern Bridge, only a locally hosted one. You'll always be "online" in the sense of needing an internet connection and talking directly to the chain.

~~Also unsure why we don't just refer to the usage instructions on the release page for local Bridge. Those are more likely to be accurate than the docs here. Anyone opposed to me taking the instructions here out and simply referring to the releases page?~~ lol those are still for old bridge, which _can_ actually be offline

The "Note: transactions cost money" part came a little out of nowhere, so I lightly restructured around it.

### DIFF
--- a/content/using/operations/using-bridge.md
+++ b/content/using/operations/using-bridge.md
@@ -10,13 +10,13 @@ aliases = ["/docs/getting-started/using-bridge/"]
 
 This guide assumes that you have an Urbit ID, or that you have found someone to send an Urbit ID to your Ethereum address and are looking to claim it.
 
-### Online Bridge
+### Hosted Bridge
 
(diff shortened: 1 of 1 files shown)


matildepark:
> Tried finding this in the docs repo first. Confusing.

I'm just gonna file an issue as an RFC.

Fang-:
Actually,  c722a7b probably isn't very sane? The things it describes match up with what's actually on the Bridge releases page: old Bridge, which does support fully-offline operation.

rmariani:
Thanks!

PR #345, opened 2020-01-29
templates: mobile templates use full grid

matildepark:
One of those things that irked me randomly tonight just idly thumbing through post directories specifically — the mobile layout seemed a little off-kilter toward the left, too narrow, imperfect ...

Turns out: the fallback responsive declaration was `c1-12`, when `c1-13` is what we want in order to use the full grid (alongside the usual gutters on the left and right).

Before / after:

image


urcades (approved):
This is really good, pls push!

PR #358, opened 2020-02-11
updates: fix slugs

matildepark:
Quick catch.

Slugs should be `post-name` not `full-section-path/post-name`, which was making some weird URL permalinks.

Also the latest post was dated 2029, so I revised that too.


urcades (approved):
lgtm

PR #361, opened 2020-02-15
urbit.org: change search hotkey, add hint

matildepark:
- Search hotkey is now `Ctrl+/` and `⌘+/`.
- Search hotkey hint added and pinned beside the search button (immediately beside in regular layout, on the end of the nav block in Docs). 

While we could make the hint switch between Ctrl or ⌘ depending on user agent, the JS always loads about a second too late — making every single page load have a blank hint suddenly pop in. Not really ideal behaviour. UX suggestions welcome, but right now I just keep it saying Ctrl, which is always going to be correct.

Screen Shot 2020-02-15 at 3 40 53 PM

image

### DIFF
--- a/sass/_main.scss
+++ b/sass/_main.scss
@@ -314,10 +314,6 @@ nav.previous, nav.next {
     // min-width: 110px;
     text-align: left;
     background: transparent;
-    span#search-icon {
-      display: inline-block;
-      transform: translate(-4px, 2px) rotate(-45deg);
-    }
     span.search-slash {
       font-size: 0.75em;
       position: absolute;
--- a/static/scripts.js
+++ b/static/scripts.js
@@ -15,8 +15,11 @@ function initHotKeys() {
       document.body.classList.remove('has-active-search-window');
     }
 
-    // / to open toggle window
-    if (key === '/' || key === 'Period' || key === 58) {
+    // Ctrl or Meta + / to open toggle window
+    if (
+      (event.getModifierState("Control") || event.getModifierState("Meta")) &&
+      (key === '/' || key === 'Period' || key === 58)
+      ) {
       if (bodyEl.classList.contains('has-active-search-window')) {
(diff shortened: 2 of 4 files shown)


urcades:
@matildepark 

Two notes:

While checking out the various pages that have search enabled on them, and seeing the screenshots above, I realize I am deeply bothered by the non-uniformity of the search inputs and where their prompts might be placed:

normalize-margins

I think in all cases, I'd much prefer we stick with an approach you've outlined in your first image, where the prompt/label is located directly to the right of the "initiate search button".

I'm not sure if this next request fits into this PR, but in playing around with how this could resolve, I found I actually preferred removing the search icon, and replacing it with directive text, such as:

`Search urbit.org` 

as a generic case or, if we wanted to get fancy:

`Search <major-page-container-here`, where what goes in between the brackets could be:

`Search Docs` or `Search Blog` or `Search Updates`, etc. (the search would of course still be universal, across all of urbit.org, so maybe the generic text is better in this regard)

Design example, with both filled button and outlined button:

small-button

I think I marginally prefer the stroked option when it's all "greyMid":

Screen Shot 2020-02-16 at 2 26 15 PM

urcades:
@matildepark let me know what you think of the above design suggestions, or whether they have a place in this PR — if you think we should split them out into another set of work, you can go ahead and merge because this PR functionally meets the requirements of the PR.

urcades (approved):

matildepark:
Design we are proud of and willing to ship is part of the PR. 

I’ll make it consistent and add logic for container directives, remove the icon, etc.

urcades:
Wonderful! This is exciting to hear – this is going to have a huge impact on people’s ability to parse docs, I believe.

urcades (approved):
Yeah this is amazing, feels much nicer than what's up now — let's merge it.

PR #362, opened 2020-02-18
Add mb4 to search

matildepark:
Search box is too close to the start of the intra-page navigation of `page_indiced` templates (Developer's Guide, FAQ, etc) — this adds some room for those cases.

### DIFF
--- a/templates/partials/navigation-search.html
+++ b/templates/partials/navigation-search.html
@@ -23,7 +23,7 @@
   {% endif %}
 {% endif %}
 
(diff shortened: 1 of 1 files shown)


urcades (changes requested):
I futzed around with trying `mb3` and that actually works perfectly I think. `mb4` was a bit too spaced out - enough to make the composition of this area feel uneven.

urcades (approved):

urcades:
All is good

PR #363, opened 2020-02-19
understanding-urbit: measure-wide

matildepark:
Understanding Urbit uses the full width of the grid — unlike all other written content on the site. This commit enforces the same `measure-wide` width across all UU elements.

image


urcades (approved):
I don't actually need to check this out to approve — let's go w/ consistency! (my bad)

PR #364, opened 2020-02-19
understanding-urbit: c4-12-lg

matildepark:
After #363, I forgot that UU uses `c4-10-lg`, not `c4-12-lg` — so they got smaller than other essays on less wide viewports within `lg`, up to the max-width of `measure-wide`. This changes that.


urcades (approved):
Yes

PR #365, opened 2020-02-19
various: all templates use c4-12-lg, measure-wide

matildepark:
All content should now be consistent I think?


urcades (approved):
Checked out all major pages, grid is indeed consistent!

PR #367, opened 2020-02-24
adds commands to check life and rift number

drbeefsupreme:
Edits to using/operations/using-your-ship.md. This information is duplicated in
https://github.com/urbit/docs/pull/812 and I'm not entirely sure it should be
there, but it should definitely be here.

This page really needs some work on re-organizing, and we (docs) have discussed
this before but not actually done anything. I think just adding a table of
contents would be a decent bandaid that will only take a few minutes, so I'll
put up another PR for that soon.

----

#

### DIFF
--- a/content/using/operations/using-your-ship.md
+++ b/content/using/operations/using-your-ship.md
@@ -182,6 +182,13 @@ Because Urbit networking is stateful we call this a _continuity breach_. Everyth
 
 When this happens, back up any files you'd like to save, shut down your urbit, and recreate it (as if you were starting for the first time).
 
+### Life and rift number
+
+You can check your ship's _life_ and _rift_ number by running the `+keys`
+generator. You can inspect the life and rift number of any ship by running the
+`.^(* %j /=life=/(scot %p <ship>))` and `.^(* %j /=rift=/(scot %p <ship>))`
+commands in dojo, respectively.
+
 ## DNS proxying {#dns-proxying}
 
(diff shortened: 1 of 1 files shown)


content/using/operations/using-your-ship.md
```diff
@@ -182,6 +182,13 @@ Because Urbit networking is stateful we call this a _continuity breach_. Everyth
 
 When this happens, back up any files you'd like to save, shut down your urbit, and recreate it (as if you were starting for the first time).
 
+### Life and rift number
+
+You can check your ship's _life_ and _rift_ number by running the `+keys`
```
rmariani: `+keys` does not seem to work. I get this:

```
> +keys
%cx /~lodleb-ritrul/home/~2020.2.25..22.15.24..9a61/gen/keys/hoon
scry failed for
ford: %hood failed for /~lodleb-ritrul/home/0/gen/keys/hoon:
ford: %core on /~lodleb-ritrul/home/0/gen/keys/hoon failed:
```

Fang-: Works on my ship. And [it certainly exists](https://github.com/urbit/urbit/blob/master/pkg/arvo/gen/keys.hoon). Something is way messed up about your home desk!

(It's `+keys ~ship`, though. So `+keys our` for our life and rift.)

rmariani:
@syzygyzer, I think this should include a short explanation as to what `life` and `rift` numbers are. Maybe a link to the guide to breaches would be warranted.

PR #368, opened 2020-02-24
Adds TOC to using-your-ship

drbeefsupreme:
This is just a bandaid for this page the makes it a little more navigable but
still needs some TOC. I removed deeper entries in the TOC since that made it
rather long.

This actually has the same content, plus the TOC, as #367. I thought I was
practicing good git hygeine by making it a separate PR, but wanted to avoid a
situation where the TOC containing the new entry for life and rift number might
get accepted before the content it refers to. So this actually makes #367
redundant, but at least it meets the goal of splitting potential discussion of
these changes into two PRs. I'll try to give it a little more forethought in the future.

----

#

### DIFF
--- a/content/using/operations/using-your-ship.md
+++ b/content/using/operations/using-your-ship.md
@@ -8,6 +8,46 @@ aliases = ["docs/using/arvo-network-dns/", "docs/using/messaging/", "docs/using/
 hidetitle = "true"
 +++
 
+## Table of Contents
+
+- [Using your ship](#using-your-ship)
+  * [Shutdown](#shutdown)
+  * [Restart](#restart)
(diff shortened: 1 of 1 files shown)


matildepark:
1. Is the sidebar table of contents not enough for this? Seems really redundant to me. 
2. If we should have a table of contents, Zola provides it as a top level variable for a page as “page.toc” if I recall, so it shouldn’t be hardcoded.

drbeefsupreme:
You're totally right, I forgot about the sidebar TOC. I was just rushing through without checking the site first since it was close to the end of the day for me, which is a silly reason to not make standard sanity checks.

PR #374, opened 2020-03-12
Revert "Update binaries to v0.10.4"

Fang-:
Reverts urbit/urbit.org#372

0.10.4 causes the OS1 OTA to not apply properly. This can be fixed by downgrading, but we should just not suggest 0.10.4 in docs for now.


vvisigoth:
LGTM

matildepark:
So when should we revert this revert?

PR #377, opened 2020-03-18
Revert "Revert "Update binaries to v0.10.4""

matildepark:
Reverts urbit/urbit.org#374 lol


Fang- (approved):
I suppose it's safe to point to latest now, yeah. Double-reverse approved!

PR #378, opened 2020-03-18
remove smooth scrolling in most contexts

matildepark:
Stops the site from making you sit and wait for it to smoothly scroll through when accessing lower content on a long page. 

Just removes smooth scrolling from the `html` tag and from the SCSS, basically; though the inter-page navigation on FAQ pages and the like still retain that behaviour since it works fine there really.


urcades (approved):
stamp!

PR #379, opened 2020-03-20
change install.md to say that chat-cli will be seen when boot finishes

botter-nidnul:
Also instruct to use Ctrl-x to switch to Dojo.

This has been a significant problem in the Urbit Discord and Urbit.live Telegram support chats.


rmariani:
A most welcome change, thanks a lot

PR #380, opened 2020-03-20
mention both dojo and chat-cli in install.md

botter-nidnul:
I've been told this can go either way: [https://github.com/urbit/urbit/issues/2248#issuecomment-601639995](https://github.com/urbit/urbit/issues/2248#issuecomment-601639995)
To cover our bases, it'd be best to mention both dojo and chat-cli as possibilities.


rmariani:
LGTM, thanks!

PR #381, opened 2020-03-20
update '~dopzod/urbit-help' to '~/~dopzod/urbit-help' since you gotta do that now

reitraced:
os1 changed the syntax for joining chats, at least on landscape. now chats that arent affiliated with a group have a ~/ at the front, which i guess denotes the group they are in internally
just gonna put this here if its okay since its really like one of the biggest mentions of u-h

it doesnt look nearly as elegant but its more correct i guess.

### DIFF
--- a/templates/index.html
+++ b/templates/index.html
@@ -153,7 +153,7 @@
     </li>
   </ul>
   <p class="lh-copy">
-    Or, boot Urbit and join <code>~dopzod/urbit-help</code>
+    Or, boot Urbit and join <code>~/~dopzod/urbit-help</code>
   </p>
 </section>
 


matildepark:
Appreciated! It also exists [here](https://github.com/urbit/urbit.org/blob/master/templates/understanding-urbit/post.html#L108) and [here](https://github.com/urbit/urbit.org/blob/master/templates/understanding-urbit/list.html#L114) and [here](https://github.com/urbit/urbit.org/blob/master/content/using/operations/using-your-ship.md) and [here](https://github.com/urbit/urbit.org/blob/master/content/understanding-urbit/people-history.md) and ... [here](https://github.com/urbit/urbit.org/blob/master/content/using/develop.md). Oh my.

PR #387, opened 2020-03-26
urbit.org: added front page banner/cta for OS1 launch event

urcades:
Something simple. This needs a proper link to a Meetup or whatever we're using to organize this event.

### DIFF
--- a/sass/_main.scss
+++ b/sass/_main.scss
@@ -76,6 +76,11 @@ nav {
     padding: 0 32px 0 32px;
 }
 
+.full-bleed-banner {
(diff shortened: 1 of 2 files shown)


matildepark (approved):

tacryt-socryp:
It's no longer launching on the 27th, right? Isn't it launching on Apr 3?

matildepark:
This is true. For reference my stamp was technically oriented lol

urcades:
Yes, I had pasted in previous info, was planning on fixing date when I get
the meeting and stream links for this thing!

—
~fabled-faster
https://urbit.org/

PR #396, opened 2020-04-08
adds a blurb on how to find the urbit community

drbeefsupreme:
Re: an urgent internal email conversation, we currently have no instructions on
how to find the Urbit Community group once you log into Landscape. I wasn't
quite sure on where to add this, but Galen said he wants a path as well-paved as
the yellow brick road, so putting it in the install doc seemed to me like the
highest-traffic lane on which to ensure people will know that Urbit Community
exists, though I think this should be seen as just a placeholder for a more
informative separate page on how to become involved in the community that we
should make soon.

I grepped for `urbit-help` in `/docs/` and `/urbit.org/` to see if there is
another spot where we already mention `urbit-help` that might be more
appropriate, but the closest thing that came up is `using-your-ship.md` and I
think a lot of people will not reach that page on day one so it didn't seem appropriate.

----

#

### DIFF
--- a/content/using/install.md
+++ b/content/using/install.md
@@ -151,3 +151,19 @@ Landscape is the Urbit web interface, and it's the best way to interact with you
 2.  If the port given is 80, simply type `localhost` into your browser's address bar. If the given port is a different number, such as `8080`, you would type `localhost:8080`. Alternatively, you access landscape by navigating to `ship.arvo.network` into your URL bar, where `sampel-palnet` is the name of your ship, without the `~`. You will need to [set up DNS proxying](@/using/operations/using-your-ship.md#dns-proxying) to use this option.
 3. Type `+code` into your ship's Dojo. Copy-paste the returned code into the field asking for it in the browser page.
 4. You're in! Now you can explore apps such as Chat for messages, Publish for blogging, and Weather.
+
+### Joining the community
+
+Once you are in Landscape we recommend joining the Urbit Community group. You
+will need to obtain an invite, which can be had by first joining
+the `~/~dopzod/urbit-help` chat channel and asking for an invitation from there.
+Once you have the invite you can accept it by navigating to your Groups by
+clicking on All Groups in the upper left hand corner of Landscape, then Manage
+All Groups, and in there you should see an invitation. From this interface you
+can join the various Chat, Publish, and Link instances contained in the group.
+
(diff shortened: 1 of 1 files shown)


drbeefsupreme:
We've have a lot of discussion on the docs team in the past about making this page as short as possible, so I don't see this as a permanent placement for this, just enough to get us by until we have a better way of joining Urbit Community than just telling people to ask for an invite.

matildepark (approved):

urcades (approved):
These instructions are looking good

content/using/install.md
```diff
@@ -151,3 +151,19 @@ Landscape is the Urbit web interface, and it's the best way to interact with you
 2.  If the port given is 80, simply type `localhost` into your browser's address bar. If the given port is a different number, such as `8080`, you would type `localhost:8080`. Alternatively, you access landscape by navigating to `ship.arvo.network` into your URL bar, where `sampel-palnet` is the name of your ship, without the `~`. You will need to [set up DNS proxying](@/using/operations/using-your-ship.md#dns-proxying) to use this option.
 3. Type `+code` into your ship's Dojo. Copy-paste the returned code into the field asking for it in the browser page.
 4. You're in! Now you can explore apps such as Chat for messages, Publish for blogging, and Weather.
+
+### Joining the community
+
+Once you are in Landscape we recommend joining the Urbit Community group. You
+will need to obtain an invite, which can be had by first joining
+the `~/~dopzod/urbit-help` chat channel and asking for an invitation from there.
```
urcades, line 159: Wonder if we should leave a disclaimer here that it's more likely to have a request fulfilled during West Coast US hours

urcades: Correction: 

> Wonder if we should leave a disclaimer here [Line 159] that one is most likely to have an invite request fulfilled during West Coast US hours.

Fang-: idk, I'm generally on during UTC hours (and can fulfill invite requests), but I'm also just one person instead of the army we have on US hours. (Generally opposed to doing and/or signalling timezone favoritism, but no way around the fact of the situation...)

urcades: Yeah, I wasn't forgetting you – just wanted to be sure we weren't putting a huge load of responsibility on a single individual or placing that expectation or making that promise, no need to warn about this if you don't feel it's necessary

PR #400, opened 2020-04-11
Fix missing word on urbit-os.md

bmaczka:
Currently reads "and we don’t think can happen using existing technology."

Updated to read "and we don’t think that can happen using existing technology."


galenwp:
This is a great catch. Thank you!

PR #404, opened 2020-04-17
various: add 'inter' fallback

matildepark:
Some kind of weirdness is going on with the Inter typeface and UI elements. 

Screen Shot 2020-04-16 at 10 45 57 PM

I added "Inter" as a fallback font for UI elements and changed the escaped HTML entity to its direct Unicode character.

Screen Shot 2020-04-16 at 10 45 48 PM

(Probably worth comparing the two in each browser.)

### DIFF
--- a/sass/_main.scss
+++ b/sass/_main.scss
@@ -145,7 +145,7 @@ footer nav {
         position: absolute;
         right: -36px;
         top: 13px;
-        font-family: "Inter UI", sans-serif;
+        font-family: "Inter UI", "Inter", sans-serif;
     }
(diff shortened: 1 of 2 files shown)


urcades:
Damn, yeah I'm encountering what's likely the same thing: 

Chrome:
Screen Shot 2020-04-16 at 11 07 56 PM

Safari:
Screen Shot 2020-04-16 at 11 09 04 PM

Very weird! Checking out fallbacks now.

urcades (approved):
Fallback is looking good in this branch, an all browsers in my machine! Stamp

PR #409, opened 2020-05-01
Minor copyedits to Introducing OS 1

mhartl:
The biggest diff is clarifying the meaning of “ship”, which in the original post is used without explanation.


galenwp:
Not sure why your editor wants to remove all those spaces — but otherwise LGTM!

Thanks!

galenwp (approved):

mhartl:
I have my editor configured to automatically trim trailing whitespace, mainly to prevent such irrelevant whitespace diffs (which otherwise inevitably creep in). Of course, unless other collaborators use the same convention, it creates the very problem it’s designed to solve! (Which convention to use can be a [bone of contention](https://github.com/atom/whitespace/issues/10).)

In future, I’ll plan to use a different editor for making changes to urbit.org, one configured not to trim such whitespace.

matildepark:
It's fine, I use the same whitespace trim.

PR #411, opened 2020-05-05
running urbit as a daemon with tmux

drbeefsupreme:
There's been some interest on urbit-help on running Urbit as a daemon and being
able to connect remotely. I've written a short tutorial on how to do so,
specifically using tmux, but I suggest GNU Screen and dtach as well (I've never
used them before).

I actually can't say for certain that Urbit cannot already be run natively as a
daemon without using a terminal multiplexer. I don't know much about this kind
of thing.

----

#

### DIFF
--- a/content/using/operations/using-your-ship.md
+++ b/content/using/operations/using-your-ship.md
@@ -213,6 +213,28 @@ You can check your ship's _life_ and _rift_ number by running `+keys our` in
 dojo. You can inspect another ship's life and rift number by running `+keys
 ~sampel-palnet`. For information on what life and rift are, see [Guide to Breaches](@/docs/tutorials/guide-to-breaches.md).
 
+## Running urbit as a daemon
+
+If you'd like to run Urbit in the background on some device and be able to
+summon it from anywhere via `ssh`, you should run Urbit as a daemon.
+
+To do this, you need a program that will keep a terminal session running even if
+you close out of it, called a terminal multiplexer. Some options are tmux, GNU
+Screen, and dtach.
+
+Here we explain how to use tmux to run Urbit as a daemon. On the machine you
+intend to run your ship, install tmux. Then begin a new session with `tmux new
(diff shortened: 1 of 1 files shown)


drbeefsupreme:
Forgot to comment on placement. `using-your-ship.md` seemed like the most straightforward place to put it, but also suffers from the problem of the page being an abomination. This should actually go into the new Knowledge Base too, but I just decided to throw this together at the end of my day and will think about that tomorrow.

galenwp:
@benjamin-tlon do we get actual daemon mode with king haskell?

baudtack:
I think that discussion of tmux/screen/etc in our documentation should be limited to mentioning they can be used. Full documentation of using any of these is well beyond the scope of the project. We don't want to be in the situation where we are updating this for whatever the flavor of the week of screen multiplexer is.

If and when we have an actual daemon option as part of urbit proper, that should be documented. My main concern here is where does the line get drawn for what is worth documenting and what isn't. Do we document bash? Do we document nix? What about make? What happens when tmux gets updated? Who is going to be in charge of making sure that every tmux update doesn't break these instructions?

galenwp:
I've been using screen in the *exact* same way (with Urbit) for the ~7 years I've worked on this project. It seems perfectly fine to document how to use our preferred options here since they're incredibly stable pieces of software. Otherwise we send people down the rabbit hole of figuring these things out for themselves. 

This is ultimately up to you all to decide — just my .2c. Actual daemon mode is much preferred.

drbeefsupreme:
Per discussion with docs team just now, I split the difference and removed tmux-specific instructions. It just suggests a few programs that you can do this with now.

I expect I'll return to this soon when King Haskell is released.

drbeefsupreme:
I actually thought this had been merged, but anyways, since this PR has been made, daemon instructions have made their way into install.md, so this is no longer relevant.

PR #413, opened 2020-05-06
Make light copyedits for communities post

mhartl:
I rolled minor edits for a few other posts in as well.


galenwp:
Thanks! These all look good to me. I'm kind of embarrassed we didn't catch them!

PR #416, opened 2020-05-10
Make light copyedits to 'Platform decay' post

mhartl:
This PR includes various minor corrections. One hard-to-see diff of note: regular hyphens (`-`) have been changed to en-dashes (`–`) in a few places. 

The main systematic change proposed by this PR is “which -> that” in phrases like “large advertising campaigns which seek”. Generally speaking, “which” should be used for dependent clauses and “that” should be used for independent clauses:

>large advertising campaigns, which seek to convert

or

>large advertising campaigns that seek to convert

but not

>large advertising campaigns which seek to convert

Hope this helps!


crrncllns:
Thanks Michael! Always good to have your input.

PR #420, opened 2020-05-13
blog: add periods to descriptions

matildepark:
Most of our post descriptions are full sentences, not dangling lines. This just lines up the two that are more loose.


urcades (approved):
Excellent

PR #426, opened 2020-05-17
Remove horizontal scrollbar on chrome and firefox

neauoire:
If you use `100vw` with padding, you will get overflow on body.

## Before

Screenshot from 2020-05-17 20-40-26@2x

## After

Screenshot from 2020-05-17 20-40-19@2x

### DIFF
--- a/sass/_main.scss
+++ b/sass/_main.scss
@@ -627,7 +627,7 @@ hr {
 
 .vid-insert-wrap {
   position: relative;
-  width: 100vw;
+  width: calc(100vw - 12px);
   height: 100vh;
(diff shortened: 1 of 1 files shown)


matildepark:
Cannot replicate on Brave or Firefox. All scale perfectly — is this ... Linux specific?

matildepark:
This PR also produces a gap on the right for me:

image

matildepark:
Confirm that this doesn't persist for you post-#428 and maybe we can just close this?

neauoire:
That fixed it :)

PR #427, opened 2020-05-17
Make clear that swap files on AWS are a bad idea

botter-nidnul:
Adding a swap file on an Amazon EBS file system will allow Urbit to boot, but once enough memory is used to cause that swap space to actually be accessed, Urbit will slow down considerably and then `<<behn>>` permanently. If swap is to be used on AWS, it should be from an Instance Store volume.

This change mentions that swap doesn't work well on EBS, and points to Amazon's documentation on how to use swap on Instance Store.

This change still might not be enough to avoid user hassles, since the most common type of EC2 instances are unable to use Instance Store. But going into a long explanation about what EC2 instances will work with Urbit and which won't seems to be beyond the scope of this install guide.


ohAitch:
I remember guidance as you needing swap because urbit will map in the whole
31bit virtual address space at once, but 1GB being entirely sufficient for
operations your pier actually performs: like, tests were run with 100ms
simulated disk seek latency, and there was no %behn to speak of. Do we know
what changed?

botter-nidnul:
Urbit is using more memory, would be my guess. From what I'm hearing from the Discord, it seems problems start around 700mg of memory usage (could just set swappiness to 0 to help with that) , and the problem becomes severe over 900mb (swap is going to have to be used around this point regardless of swappiness).

OTAs seem a likely culprit for driving Urbit into swap territory? Maybe people are experiencing the kernel using lots of memory? I've seen several logs posted in issues on github.com/urbit/urbit where `|mass` is reporting a kernel size of over 500MB, or even 1GB.

cxk280:
This PR would have saved me some heartache if it had been merged. You might want to prioritize it.

matildepark:
Aye, I just don't know who to assign this to now. @syzygyzer? @baudtack? One of you want to handle this?

(And @botter-nidnul, could you resolve the merge conflict?)

cxk280:
Also, I do think some general guidance on EC2 instance types should indeed be in the scope of the install guide. If not suggesting specific instance types, at least characteristics to look for. The default instance types are t2, which is EBS-only storage. Sounds like that's something to avoid.

cxk280:
It looks like `m5ad.large`, `c6gd.medium`, and `c6gd.large` are some of the cheaper types with Instance Stores.

botter-nidnul:
let me try that again...

botter-nidnul:
I don't understand what's happening here. I cloned the master branch of urbit.org, copied install.md into my branch, added my changes, and it's detecting conflicting changes that aren't conflicting (that are in fact exactly the same, because it's the same file)

Do I need to delete the branch and create it again? Squash commits?

matildepark:
I resolved it.

baudtack (approved):

drbeefsupreme (commented):
lgtm

botter-nidnul:
I spelled `accessed` as `accesed` 😳

PR #428, opened 2020-05-18
Fix video causing overflow

oskarkunik:
This should fix horizontal overflow of the page's body.

Tested on latest FF and Chromium, for desktop and mobile resolutions.

### DIFF
--- a/sass/_main.scss
+++ b/sass/_main.scss
@@ -627,7 +627,7 @@ hr {
 
 .vid-insert-wrap {
   position: relative;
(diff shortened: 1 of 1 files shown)


matildepark:
I ... do not have any horizontal overflow, and I'm wondering why this is.

matildepark:
This doesn't seem to introduce any issues, and seems to fix it for those who encounter this, so I'll just let this in.

neauoire:
Fixes it for me

PR #434, opened 2020-06-03
new binary instructions

drbeefsupreme:
Adds instructions to `using/install.md` on how to update to the latest binary.

One thing I am unsure about - does your ship need to be mounted before moving
its pier to the folder with the new binary? I will edit it to say as much if so.
I put it below the instructions for mounting thinking that this might be the
case, but if not then I would consider moving it higher on the page.

----

#

### DIFF
--- a/content/using/install.md
+++ b/content/using/install.md
@@ -157,11 +157,12 @@ Good, your Dojo is working correctly. Now let's run our first useful command in 
 
 The `>=` output means that a command was successful. Now you can see your ship's files in its Unix directory.
 
-## Updating to the binary
+## Updating to the latest binary
 
 Most updates to Urbit are downloaded and applied automatically as OTA (Over the
 Air) updates. Occasionally it would be infeasible to distribute an update this
-way, and a new `urbit` binary is released. To update to the latest binary, follow
+way, and a new `urbit` binary is released. This is announced in the `urbit-dev`
+Google Group when it occurs (as are all OTA updates). To update to the latest binary, follow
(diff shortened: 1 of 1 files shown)


joemfb:
Your pier can be anywhere, there's no constraint whatsoever that it be in the same directory as the binary. I can't think of anything that's gained by recommending this pattern -- what's the motivation?

drbeefsupreme:
@joemfb I was unaware of this. I had assumed that was necessary since that is where the pier first appears when you make a new ship. How can you put your pier somewhere other than the directory the binary is in?

rmariani:
@joemfb I've seen a lot of people get confused with pier/binary location. I think this pattern makes sense for n00bs, who are presumably most of the people scratching their heads on `install.md`

ohAitch:
Is it the same directory? I thought the binary would be in bin/urbit

drbeefsupreme:
Sure if you move it there, but by default the binary is in the folder that comes out of the .tgz file and we don't have anything on the install page that suggests moving it from there. And from experience with helping users with pier problems, it seems most people do not move it.

matildepark:
> How can you put your pier somewhere other than the directory the binary is in?

Say you're in the folder for the urbit binary in your home folder (eg. `~/urbit-v0.10.4-darwin`), but your pier is in your home folder (`~/sampel-palnet`).

From within your Urbit's folder, you'd enter `./urbit ../sampel-palnet`.

drbeefsupreme:
Given the new shit that has come to light, I'll make another PR tomorrow addressing this stuff.

PR #435, opened 2020-06-04
new binary instructions revise (#434)

drbeefsupreme:
This is a followup on #434, which was merged a bit hastily.

This adds clarifying language on what the argument in `./urbit sampel-palnet`
actually is (the path to the pier rather than the name of the ship) and how it
has relevance to updating to the latest binary.

I still have a lingering question from yesterday, which is confirmination that
you do not need to mount the ship's filesystem in Unix to move the pier. I did a
couple tests with fake zods to confirm that you do not need to mount, but it leaves me confused about
where the state of a ship is actually stored before it has been mounted, since
everything in Unix is supposed to be a file. That's a separate question from this PR, but I
still wanted to mention it in case I was lead to the wrong conclusion from my tests.

----

#

### DIFF
--- a/content/using/install.md
+++ b/content/using/install.md
@@ -127,9 +127,26 @@ Either command will create a directory called `sampel-palnet/` and begin buildin
 
 When your ship is finished booting, you will see either the `~sampel-palnet:dojo>` or `~sampel-palnet:chat-cli/` prompt. If you're seeing `:chat-cli` press `Ctrl-x` to switch into Dojo. At that point, you should permanently erase your keyfile from your machine.
 
-To shut down your ship, use `Crtl-d`. 
+To shut down your ship, use `Ctrl-d`. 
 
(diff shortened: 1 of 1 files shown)


content/using/install.md
```diff
@@ -127,9 +127,26 @@ Either command will create a directory called `sampel-palnet/` and begin buildin
...
+can put it wherever you like.
+
+Never boot multiple instances of your ship at the same time. You can prevent
+this from happening on accident by only ever keeping a single copy of your pier.
+
+### Updating to the latest binary
+
+Most updates to Urbit are downloaded and applied automatically as OTA (Over the
+Air) updates. Occasionally it would be infeasible to distribute an update this
+way, and a new `urbit` binary is released. This is announced in the `urbit-dev`
```
rmariani: you should probably link the urbit-dev group when you mention it : https://groups.google.com/a/urbit.org/forum/#!forum/dev

PR #454, opened 2020-07-23
various: point users to urbit community

matildepark:
- Sig prepended paths are dead
- We can now point people directly to Urbit Community, so this changes copy likewise
- Some intro copy is amended for a flow that directs people to the Group paradigm first instead of the IRC-style flow


rmariani (approved):
Looks good, thanks for this update

PR #460, opened 2020-08-11
adding -p ames_port to local install instructions

dolled-possum:
This commit adds some detail about using -p <ames_port> to the instructions for running a ship locally, as well as indicating the need for UDP port forwarding.

I'm not thrilled with my update here, but I thought it was still worth it to get something out there about it.  Still I think it might be improved by a little bit of restructuring of the current instructions to lay out the importance of being able to reach the ship's ames_port, why it's less relevant a hosting scenario, how firewalls may also impact this, how to test to make sure your port is patent, etc.  One step at a time, though.

### DIFF
--- a/content/using/install.md
+++ b/content/using/install.md
@@ -351,24 +351,26 @@ Find the absolute path to the keyfile that you downloaded from Bridge. Copy it.
 
 If you are not already within the directory you installed above, enter it by running `cd urbit-v0.10.8-darwin` (for Mac) or `cd urbit-v0.10.8-linux64` (for Linux) from where you ran the install commands. It contains your Urbit binary, and your ship will be installed here as well.
 
-Once you're inside, run the command below, except with `sampel-palnet` replaced by the name of your
+For other ships on the network to communicate with your ship, the computer it's running on needs to be able to receive inbound UDP traffic.  Therefore, it's important when running your ship locally to use the `-p ames_port` flag (replace `ames_port` with a public port number from 49152 to 65535) to specify on which port you'll receive UDP traffic, and to forward inbound UDP traffic arriving at your router on that port to your computer.
+
+Once you're inside, run the command below, except with `ames_port` replaced with a public port number, `sampel-palnet` replaced by the name of your
 Urbit identity, and `path/to/my-planet.key` replaced with the path to your keyfile:
 
 ```sh
-./urbit -w sampel-palnet -k path/to/my-planet.key
+./urbit -p ames_port -w sampel-palnet -k path/to/my-planet.key
 ```
 
 Or, if you'd prefer to copy your key in, you can run:
 
 ```sh
-./urbit -w sampel-palnet -G rAnDoMkEy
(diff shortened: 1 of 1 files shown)


drbeefsupreme:
It's important to keep the install page as minimal and simple as possible. I don't have any actual hard evidence, but I don't think most people need this flag simply based on the fact that I've very rarely seen it mentioned on Landscape. It should definitely be documented, but I think would be more at home on the Using Your Ship page.

Semi-related issue: https://github.com/urbit/docs/issues/919

botter-nidnul:
> I don't think most people need this flag simply based on the fact that I've very rarely seen it mentioned on Landscape. 

That's just because I grind my teeth every time I see someone in the urbit-community channels who has trouble joining groups, instead of posting instructions about binding and forwarding the ames port. I've become convinced that a large proportion of the "I can't join X" complaints I see there are due to this issue, somehow (a timeout being reached because of the slowness?), and I've seen these problems immediately resolve when the ames port is properly exposed. I think this is a **massive problem**, for almost _every_ locally hosted Urbit - but it's currently not easy to diagnose or fix.

natpmp with urbit-king should help with this a bit, but there needs to be messaging within urbit about whether or not the ames port is externally reachable: a message on each restart of the pier next to `ames: live on XXXXX`, and maybe a line added to `+trouble`. Currently, the only way I know to help someone diagnose this problem is to ask them to run something like `=/(ss .^(ship-state:ames ax+/=//=/peers/~litzod) ?>(?=(%known -.ss) route.ss))`.

joemfb:
@botter-nidnul, at the limit, I think something like that is the only way to diagnose this problem -- "am I reachable via a public route" is an empirical question, only answerable from a perspective other than you're own.

I agree that `-p` should be standard advice on cloud servers. But I've never used it with my planet, always running locally across a wide variety of ISPs, for several years, with apparently decent %ames performance.

joemfb:
Which is not to say that we shouldn't build in some automatic detection and messaging around indirect routes -- it's just that such a feature will involve something like communicating with your sponsor to get their perspective.

drbeefsupreme:
Alright so I've been doing some more poking around with how Urbit uses ports and now understand the importance of publicizing this command a bit better. The install page is as far as many people will get for a long time, and port issues can create some serious problems with chat sluggishness that could easily convince someone that Urbit performs too poorly to be usable.

So now I do believe in saying something about this on the install page. I'm not sure exactly what yet, but I wanted to comment on this now so that you're aware that I'm informed enough now to review this PR and will be coming back to it later this week.

baudtack:
@drbeefsupreme did you have any  other thoughts on this thread?

drbeefsupreme:
Sorry this has lingered for so long. I'm still of the opinion that this breaks the unwritten rule that the install docs should be as simple as possible, but clearly I have not come up with a better suggestion after all this time, so at this point I ought to just approve it.

drbeefsupreme (approved):

drbeefsupreme:
I'll deal with the merge conflict tomorrow* (he says, with painful awareness of previous broken promises)

PR #466, opened 2020-08-23
docs: add cloud hosting docs

zalberico:
I wrote detailed cloud hosting documentation on my personal blog.

As part of an in progress grant I'm moving this documentation to the official site.

I've made some minor changes to better match the Urbit site's style.

Grant: https://grants.urbit.org/proposals/1452740478
Blog: https://zalberico.com/essay/2020/06/06/urbit-on-the-cloud.html


zalberico:
@botter-nidnul for SA (I don't have permissions to add you as a reviewer in the sidebar).

I'm also having issues getting the zola static site generator to run so I haven't seen the generated html locally yet.

matildepark:
@zalberico Check the [deploy preview](https://deploy-preview-466--urbit-org.netlify.app/) here.

botter-nidnul:
The `-p 32123` argument needs to be added to the screen section. That's the only omission I see. 

More broadly though, I'm still not happy with nginx being used for this purpose and I'm not sure what the justification is. 

My concern is not just the extra steps it adds to the guide, but that it contradicts the Urbit philosophy (as I understand it). Nginx adds complexity to the running of a personal server, and increases management and maintenance difficulty. It's not just another piece of infrastructure Urbit is ~~parasitizing~~ epiphyting, nginx is sitting *above* Urbit, providing an essential function required for Urbit's basic operation (landscape), and duplicating a function that already exists in Urbit itself. 

Urbit hasn't been audited yet, but I don't think granting it the net_bind_service capability with setcap - so it can handle what nginx is doing here - is all that threatening on a simple, consumer run DO droplet. 

I think allowing Urbit access to port 80, and letting Urbit's acme handle the Let's Encrypt certificate requests, is best practice here.

baudtack (commented):
The guide here is fine for what it is. But I share the concern about using nginx in a primary guide. I also don't think having hosting instructions for a specific host is something that should be hosted on urbit.org. Having these instructions is great, but they should be on someone's blog, not living on urbit.org as an endorsement of this company. Urbit should be neutral with regard to hosting. We can suggest possible hosts, but providing full instructions is too much.

galenwp:
I think it'd be great to remove nginx as a part of this. But I disagree that something like this shouldn't be on urbit.org. It's ideal for people to host their planets in the cloud somewhere, I see no reason not to provide basic instructions on how to do so. If we had instructions for a few cloud hosts, that'd be great.

matildepark:
> Having these instructions is great, but they should be on someone's blog, not living on urbit.org as an endorsement of this company.

I mean, Tlon.io represents the Tlon Corporation. Urbit.org is about the project; and hosting a ship is a basic documentation feature for the project specifically. Community docs and content will become more and more common ...

matildepark:
Apparently Cmd+shift+enter closes an issue.

zalberico:
Thanks for all of the feedback everyone :).

I'll refactor and remove the Nginx steps from the guide in favor of Urbit DNS arvo config (@botter-nidnul has helpfully written some of this up in a PR to my personal blog).

I'll also second the comment that it's important to have cloud hosting docs like this for new users. Getting a hosted urbit in the cloud is critical for new users being able to use urbit on multiple devices and a supported document we know works that makes things easier for them (and is kept up to date) is critical for removing that barrier.

I think over time this may become less true when there's an automated hosting option (Tlon is working on this) that is similar to what Matrix Element hosting is, but for now I think there's a lot of value in having explicit steps to guide the user and not forcing them to search the internet to find outdated or insecure setup instructions on their own.

[Edit: I think a note stating that urbit.org doesn't endorse any specific provider or company is also reasonable. My long term preference would be steps for the main providers, DO, AWS, Gcloud, Azure, maybe some home server raspberry pi option or something. I think getting urbit deployed on a server is a core part of the user experience and should be owned for it to be good]

baudtack:
Is urbit.org going to maintain this document then? Does it have a droplet that this can be tested on? Is someone going to test this regularly to make sure these are still up to date?

matildepark:
Urbit.org is being born, but I see it going that way, yeah. It's good to have a central point for editing as opposed to a million single-repo guides from years past.

zalberico:
Ultimately yes, when it breaks we should fix it (and users will let us know).

baudtack:
I just don't think it's the responsibility of urbit.org to teach people to use linux and we should be extremely aggressive in descoping things to avoid having a maintenance nightmare. This seems to be more a discussion of what the nature of urbit.org is and probably best for another place and with those working on it involved.

zalberico:
I agree it's not urbit.org's responsibility to teach linux, but there is a responsibility to make urbit deployment easy for user adoption. 

I'm not sure a better path to get there other than explicit steps.

Ack on concern about scope expansion (definitely common problem/risk you're right to push back on), but I think one hosting example (even if it's only one) is helpful for making things way easier for users.

I'd argue not doing this actually incurs some future risk where centralized deployment services that abstract complexity away from the users entirely could gain ground.

[Edit: grammar]

matildepark:
@jalehman opinions?

jalehman:
> Urbit.org is about the project; and hosting a ship is a basic documentation feature for the project specifically. Community docs and content will become more and more common ...

I agree with this. Urbit.org should host a wide range of content that helps newcomers get set up, which means that it'll need to accommodate a wide range of different setups. A single end-to-end hosting example like this one has the potential to bring someone onto Urbit that wants to self-host, but is otherwise not proficient enough with Linux to get up and running.

> I just don't think it's the responsibility of urbit.org to teach people to use linux

It's definitely not our responsibility to teach anyone Linux, but as long as Linux is the only (edit: it's not the only one, but it is the *most* viable one) viable option to self-host your ship, the least we can do is, as a trusted authority, get someone past the Linux bits by telling them what to do to safely use Urbit in that environment.

> avoid having a maintenance nightmare

While having a maintenance nightmare is possible, it's not a foregone conclusion that one will occur if we expand the scope of the existing documentation. It's also possible that the number of contributors also expands and fills keeps this documentation updated as mistakes are encountered. If such a maintenance nightmare does occur, that's okay—I'm confident that we can find a path forward from there. We might then look to offload some of the more platform-specific bits to a platform that better supports user-generated content (e.g. a Wiki), rework the structure of urbit.org, or just delete documentation that is inaccurate and unmaintained.

Until we arrive at a maintenance nightmare though, I'd err on the side of admitting more documentation that stands to ease the onramp to Urbit.

zalberico:
I've removed the nginx config. I also noticed that @botter-nidnul has put some work into integrating some DO docs already here: https://github.com/urbit/urbit.org/commit/0ec1521b303e3d75855cf48f5f80d6e8007d9a9b

@botter-nidnul - what do you think the right approach is? - it might make that install page simpler to break out the cloud hosting into its own section as I've done here and then just link to it from the install page. This page also goes into more of the user/domain/DNS setup which I think is useful to have in the core guide. Happy to merge these together or include anything from the work you've already done that you think I'm missing.

zalberico:
I'm going to take a crack at simplifying the existing install docs (which have gotten a little unwieldy) as part of this PR.

Basic idea is to break out the hosting bits from the install page, will be clearer once I push up the changes.

zalberico:
Apologies about the messy history, but I think things are finally in a good place to be reviewed.
I just reread it myself, but it'd be helpful to have another set of eyes go through it.

Main highlights:
- Link to the hosting document for cloud hosting, remove those steps from the install guide.
- Reduce a lot of repeated install instructions (makes the page more concise, reduces maintenance required for keeping docs up to date)
- Delete separate comet doc and put the handful of steps on install page (they were previously in both locations)
- Remove 'Make a Note' in favor of just having chat, once in chat users can figure out how to make a note and I don't think new users need that extra information in the install guide.
- Delete 'supernode' in favor just saying star or galaxy.
- Overall bias towards being concise as possible *while maintaining clarity*. 
- Fix typos

I think this is good to go, but feel free to comment on the changes.

matildepark:
*pokes @baudtack *

content/using/install.md
```diff
@@ -149,40 +159,32 @@ Click that `Chat` type on your Landscape home page, and then click the `Join Cha
...
-### Getting Urbit ID
+## Getting an Urbit ID (Planet or Star)
 
 Tlon doesn't currently sell or distribute Urbit IDs, but there are a few ways to get your own:
 
 - Getting an invitation from a friend
 - Purchasing an Urbit ID from a third party, such as [urbit.live](https://urbit.live), [OpenSea](https://opensea.io), [planet.market](https://planet.market/), or [Urbit Marketplace](https://urbitmarketplace.com/).
 
+Note that if you buy a planet from opensea you may have to change its star, ask around in the `urbit-community` chat as a comet first for help before purchasing.
+
```
jalehman: This is a good opportunity to link to the section about escaping your sponsor.

content/using/install.md
```diff
@@ -149,40 +159,32 @@ Click that `Chat` type on your Landscape home page, and then click the `Join Cha
...
 
-- **Master Ticket**: The "password" at the top of the hierarchy that determines the ownership. All other secrets are derived from this. If the rest of the wallet is lost or compromised, the master ticket can rederive everything. You'll log into Bridge with this. Always keep your master ticket extremely safe.
+- **Master Ticket**: The "password" at the top of the hierarchy that determines the ownership. All other secrets are derived from this. If the rest of the wallet is lost or compromised, the master ticket can derive everything. You'll log into Bridge with this. Always keep your master ticket extremely safe.
 
 - **Management Proxy**: Derived from the master ticket. Can perform non-ownership related operations such as configuring Urbit OS networking keys.
 
-- **Keyfile**: Derived from the management proxy. Used as crytographic proof that your Urbit ship it is who it says it is. You use it to start up your ship for the first time.
+- **Keyfile**: Derived from the management proxy. Used as cryptographic proof that your Urbit ship it is who it says it is. You use it to start up your ship for the first time.
```
jalehman:
```suggestion
- **Keyfile**: Derived from the management proxy. Used as cryptographic proof that your Urbit ship is who it says it is. You use it to start up your ship for the first time.
```

zalberico:
Once this merges, I'd also like to change the planet boot instructions so the ship is not inside the binary directory. This will make upgrading the binaries a lot simpler and less confusing because the pier will be separate.

Users can swap out the binary dir and it'll be clear that they can reuse their existing pier directory (I'll change the docs to do things this way by default). This should make binary upgrades less error prone for users.

I'd prefer to merge this PR first though since it's is already quite long and hard to review.

content/using/install.md
```diff
@@ -193,151 +195,17 @@ As mentioned previously, there are a few ways to acquire a planet. All methods,
...
+There are two supported ways of running an Urbit ship: using a cloud service, or running it locally on your own machine. Most users run things locally at first, but we recommend using a cloud service for Urbit because it allows your ship to be accessed from anywhere on multiple devices. Hosting your ship in the cloud also allows it to always be online and ready for OTA updates. It's technically possible to run your Urbit ship on a home server, but ISPs often restrict this to business plans and opening up your home network to the internet can be unsafe if done improperly.
 
-Urbit cannot run on Windows itself, but there is a convenient way to run a Linux distro using the [Windows Subsystem for Linux 2](https://en.wikipedia.org/wiki/Windows_Subsystem_for_Linux), an initiative by Microsoft to bring Linux to Windows 10. For Urbit to work, it is necessary to [install WSL 2](https://docs.microsoft.com/en-us/windows/wsl/wsl2-install) and not just WSL, since version 2 uses virtualization and the real Linux kernel under the hood.
+Using a cloud service costs money, but it's more convenient to use than a local install.
```
jalehman: Is a cloud service more convenient? It's more of a PITA in the short-term, but has all of the benefits listed above. I'm not sure this line is necessary.

content/using/install.md
```diff
@@ -457,11 +279,10 @@ Galaxies are rare and extremely valuable, meaning that there is no liquid market
...
 
-- Printing the secret on a piece of paper. However, paper wallets are vulnerable to various forms of physical damage, such as rot, water damage, smoke, or fire. Laminating the paper can mitigates some of these risks, but the lamination can potentially trap moisture. Make sure that you trust the printer: some have memory and network connections.
-- Storing the secret on a brand-new USB stick or hardware wallet (such as a Ledger) that is never connected to a networked machine.
-- Engraving the secret on a strip of stainless steel. This medium is resistant to both water damage and fire damage.
+- Storing the secret on a hardware wallet (such as a Trezor or Ledger) that is never connected to a networked machine.
+- There are products that let you store the key on steel (such as Cryptosteel or BILLFODL). This medium is resistant to both water damage and fire damage.
```
jalehman: It seems like there's some redundancy here now.

content/using/install.md
```diff
@@ -473,79 +294,9 @@ It’s a good idea to store your keys redundantly; for example, on both a USB st
...
-
-Your ship should be booting, and displaying a series of boot messages starting with `urbit 0.10.8`. A directory called `sampel/` and begin building your ship. It may take a few minutes.
-
-When your ship is finished booting, you will see either the `~sampel:dojo>` or `~sampel:chat-cli/` prompt. If you're seeing `:chat-cli` press `Ctrl-x` to switch into Dojo. At that point, you should permanently erase your keyfile from your machine.
-
-To shut down your ship, use `Ctrl-d`. Since you're running, you won't have much reason to do this often, since you should always have it running in your Droplet. More frequently you'll want to use `Ctrl-a` and then `d` to go back to your Droplet's command line while keeping your planet running on a detached Screen. Then, to return to your ship, run `screen -r`.
-
-If you need to shut down your ship, start it up again by running `./urbit sampel` from the directory where your Urbit binary is saved. Note that in this usage, `sampel/` is the path of a folder, which by default is located in the same folder as the Urbit binary. This folder is called your ship's **pier**.
+Because of their need for near-perfect uptime Stars should use cloud services.
```
jalehman:
```suggestion
Because of their need for near-perfect uptime Stars should be hosted on [cloud services](insert link here). 
```

jalehman (commented):
A few initial thoughts.

content/using/install.md
zalberico: 👍 I was tempted to delete it, but I think it was already there in the mass of docs I removed. Will delete.

content/using/install.md
zalberico: Good catch - thanks

content/using/install.md
zalberico: On this one I didn't include the link because the next line has the link to the guide. I did switch it to your language though.

zalberico:
Thanks for the review - I've committed all the changes.

content/using/install.md
```diff
@@ -8,35 +8,29 @@ aliases = ["/docs/getting-started/"]
...
 
-But sometimes a user might not want a persistent identity for whatever reason. In that case, they can use Urbit without an Urbit ID by using something called a **comet**. Comets can be spun up at will and for free. They are good for bots and for people who are new to Urbit and want to check out the network, but they aren't good for building a reputation as a friendly and responsible peer. If you want to jump right in with a comet, check out [this](#comet) section.
+There are also free IDs meant for temporary use called **comets**. They are good for bots and people who are new to Urbit and want to check out the network, but they aren't good for building a reputation as a friendly and responsible peer. Their long names make them difficult to remember and their low default reputation makes them less likely to get over the air updates. Some communities ban comets to prevent spam, but they're still a good way to see what Urbit is about before buying your own planet.
```
baudtack: Object to saying comets are for temporary use.  Comets have many potential uses completely tangential to a sort of trial mode of urbit.

Reputation is not what prevents them from getting OTAs.

content/using/install.md
```diff
@@ -82,9 +76,25 @@ Urbit cannot run on Windows itself, but there is a convenient way to run a Linux
...
+
+### Comet: Install and Setup {#comet}
+
+If you don't have your own Planet ID, but still want to try out the Urbit network you can launch a comet.
+
+**Comets** are temporary 128-bit or 16 syllable Urbit IDs, that look like:
+
+`~dasres-ragnep-lislyt-ribpyl--mosnyx-bisdem-nidful-marzod`
+
+Comets are disposable, free identities that anyone can make to join the live network and try things out.
```
baudtack: Same objection to the phrasing here. Better might be "and can be used to try things out."

content/using/install.md
```diff
@@ -82,9 +76,25 @@ Urbit cannot run on Windows itself, but there is a convenient way to run a Linux
...
+
+### Booting a comet
 
-To boot a new comet, go into the command line and run the following command from the directory that was created during Urbit installation:
```
baudtack, line 87: extremely minor, but why was new removed?

content/using/install.md
```diff
@@ -149,40 +159,34 @@ Click that `Chat` type on your Landscape home page, and then click the `Join Cha
...
-
-### Getting Urbit ID
+## Getting an Urbit ID (Planet or Star)
 
 Tlon doesn't currently sell or distribute Urbit IDs, but there are a few ways to get your own:
 
 - Getting an invitation from a friend
 - Purchasing an Urbit ID from a third party, such as [urbit.live](https://urbit.live), [OpenSea](https://opensea.io), [planet.market](https://planet.market/), or [Urbit Marketplace](https://urbitmarketplace.com/).
 
+Note that if you buy a planet from opensea you may have to change its star, ask around in the `urbit-community` chat as a comet first for help before purchasing.
```
baudtack: Should be noted here that this possibility of having to escape comes with buying a planet from _anyone_ not just opensea. You're more likely to have a problem there as it's not an urbit specific market, but it should be made clear this is not something specific to opensea.

content/using/install.md
```diff
@@ -457,11 +279,10 @@ Galaxies are rare and extremely valuable, meaning that there is no liquid market
...
+It’s a good idea to store your keys redundantly in case one method fails. If you deem a key to be valuable enough, you can shard it into thirds and store each third in secure, geographically distributed locations. Consider these storage media options:
 
-- Printing the secret on a piece of paper. However, paper wallets are vulnerable to various forms of physical damage, such as rot, water damage, smoke, or fire. Laminating the paper can mitigates some of these risks, but the lamination can potentially trap moisture. Make sure that you trust the printer: some have memory and network connections.
-- Storing the secret on a brand-new USB stick or hardware wallet (such as a Ledger) that is never connected to a networked machine.
-- Engraving the secret on a strip of stainless steel. This medium is resistant to both water damage and fire damage.
+- Storing the secret on a hardware wallet (such as a Trezor or Ledger) that is never connected to a networked machine.
```
baudtack: Why was paper removed?

content/using/install.md
```diff
@@ -457,11 +279,10 @@ Galaxies are rare and extremely valuable, meaning that there is no liquid market
...
+- There are products that let you store the key on steel (such as Cryptosteel or BILLFODL). This medium is resistant to both water damage and fire damage.
```
baudtack: Let's not mention specific products for this.

baudtack: Honestly we should probably not for hardware wallets either.

baudtack (changes requested):
Generally looks really good. Is there a pr for the docs repo side of this?

content/using/install.md
zalberico: What prevents them from getting OTAs? I asked on chat and someone told me it was because they don't get prioritized for compute since they have low default reputation.

zalberico: I'll change the language to remove 'temporary'

content/using/install.md
zalberico: I don't think it was - I think this is a github diff bug.

content/using/install.md
zalberico: I've read paper is just generally problematic (often destroyed/lost), I can add it back if you think we should have it.

content/using/install.md
zalberico: Will remove the specific companies

zalberico:
I'll make these changes and push up a commit.

content/using/install.md
zalberico: I opted not to add paper back here (I thought of adding it with a flood/fire caveat - I've just read a lot of disaster stories of keys lost on paper), seems like if we're going to suggest things we might as well suggest the two better options.

zalberico:
Thanks for the review feedback and comments.

I pushed up the changes.

zalberico:
> Is there a pr for the docs repo side of this?

I think @matildepark took care of the changes on the docs repo side for this yesterday (I think the only needed change were the relative links in the glossary for comet).

I don't think there's anything else I need to do in the docs repo?

matildepark:
https://github.com/urbit/docs/commit/ee4c4bbbf0ea19d89781bd73b7d7bff0408fd8e7 :P

zalberico:
@baudtack if the changes look good to you, could you approve? Hoping to get this merged today. :)

baudtack (approved):
looks great!

PR #468, opened 2020-09-04
docs: add binary upgrade instructions

zalberico:
The current structure of having the pier inside of the extracted binary directory causes two problems for users.

* When upgrading the binary they have to move their ship out of the old binary directory.
* It's less obvious to users that the ship can be reused.

By structuring the docs to create a parent `urbit` directory to house both the binary directory and the ship within it as two distinct directories it's a lot easier for a user to see the structure. It also simplifies upgrading the binaries to newer versions (while reducing the chance for user error).

I also include an explicit upgrade example to help guide the user through this process.

Key Changes:
* Change install docs to create an urbit parent directory
* This urbit directory contains a diectory for the binaries and a directory for the ship
* This simplifies binary update instructions for the user, making it easy to swap out the binary directory
* Add an example binary upgrade with commands to help guide the user.
* Explicitly tell the users that existing ships can be reused.
* This structure should reduce user errors on upgrade and provides more clarity between binary and pier.

Misc Fixes:
* Remove outdated Nginx mention in hosting
* Link to breaching instructions in hosting doc
* Be consistent with sampel_commet use

### DIFF
--- a/content/using/install.md
+++ b/content/using/install.md
@@ -45,19 +45,21 @@ Urbit: a personal server operating function
 ### macOS
 
 ```sh
+mkdir urbit
+cd urbit
 curl -O https://bootstrap.urbit.org/urbit-v0.10.8-darwin.tgz
 tar xzf urbit-v0.10.8-darwin.tgz
-cd urbit-v0.10.8-darwin
-./urbit
+./urbit-v0.10.8-darwin/urbit
 ```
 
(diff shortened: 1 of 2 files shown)


zalberico:
@baudtack , @matildepark , and @botter-nidnul for SA.

Thanks!

baudtack (approved):
fantastic work

PR #469, opened 2020-09-05
docs: clarify instructions, OS indepedence

zalberico:
This is an iteration/polish improvement on my previously merged PR: https://github.com/urbit/urbit.org/pull/468

Details listed below, but core overview is that I went through and brought the planet boot instructions in line with the new directory structure used in setting up comets (to make binary upgrades easier for users). I also changed the example to just show users commands for both macOS and Linux.

I think this simplifies things for users by making things more explicit - I think showing the key deletion command (while trivial) will also make it less likely for users to glance over that comment in the paragraph and miss the step (in my experience people tend to notice code blocks more than they do sentences in text).

* Show explicit commands for both macOS and Linux where appropriate
* Bring planet boot instruction directory structure into alignment with instructions for comets
* Show explicit command for removing keyfile (harder to miss than note in paragraph)
* In binary upgrade example, remove old binary before untarring new one (prevents name collision)
* In binary upgrade example just show users explicit commands for macOS and Linux.
* Fix missing breaches tag link

### DIFF
--- a/content/using/install.md
+++ b/content/using/install.md
@@ -98,7 +98,7 @@ Comets are disposable, free identities that can be used to quickly join the netw
 
 ### Booting a comet
 
-To boot a comet, go into the command line and run the following command from the urbit directory you created during Urbit installation:
+To boot a comet, go into the command line and run the following command from the `urbit` directory you created during Urbit installation:
 
 ```sh
 macOS:
@@ -108,7 +108,7 @@ Linux:
 $ ./urbit-v0.10.8-linux64/urbit -c mycomet
 ```
 
-This will take up to an hour, and will spin out a bunch of boot messages. It will also create a directory called `mycomet` in the Unix directory that you ran the command from (which should be your urbit directory).
+This will take up to an hour, and will spin out a bunch of boot messages. It will also create a directory called `mycomet` in the Unix directory that you ran the command from (which should be your `urbit` directory).
 
 Toward the end of the boot process, you'll see something like:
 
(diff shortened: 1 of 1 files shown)


zalberico:
@baudtack, @matildepark, @botter-nidnul for SA - thanks!

content/using/install.md
```diff
@@ -118,9 +122,16 @@ http: live (insecure, loopback) on 12321
 
 When your ship is finished booting, you will see either the `~sampel_commet:dojo>` or `~sampel_commet:chat-cli/` prompt. If you're seeing `:chat-cli` press `Ctrl-x` to switch into Dojo. 
 
-To exit Urbit, use `Ctrl-d`. 
+To exit Urbit, use `control-d` or enter `|exit` into Dojo.
```
matildepark: `Ctrl+D` is actually clearest. But be consistent with the `Ctrl-x` right above it...

zalberico: I'll make the change

content/using/install.md
```diff
@@ -225,65 +236,106 @@ Find the absolute path to the keyfile that you downloaded from Bridge. Copy it.
...
-First shut down your ship, by running the following in dojo (or control-d).
+First shut down your ship, by running the following in dojo (or `control-d`).
 ```
 |exit
 ```
 
-Next download and extract the most recent binary from inside your `urbit` directory:
+Delete the old binary directory from your `urbit` directory:
+
+**NOTE**: Your ship itself can be reused, only the binary needs to be changed. This is why we have the directory structure configured as an `urbit` directory that contains your ship and your binary. This structure makes it easy to swap in new binaries. If you have your ship *inside* the `urbit-vX.XX.X` directory, move it out before deleting that directory.
```
matildepark: A little verbose. We're about to demonstrate that it's the same ship on line 327

zalberico: I'm worried people who have previously put their ship in their binary directory might delete it if they don't take it out first.

So it's a little verbose, but I thought explicitly stating ships can be reused and to not delete your old one would be good.

matildepark:
Screenshot 2020-09-06 at 1 10 51 AM
Screenshot 2020-09-06 at 1 10 55 AM

inconsistent

zalberico:
I think this is good to merge

matildepark:
lol, but the first style was better tho

zalberico:
I think the first style takes up a lot more vertical screen space without any real clarity benefit.

zalberico:
Unless you have really strong objections to this, I'd vote to merge (I think it's a nit/preference?).

Obviously if this will block merge I'll change it.

After this merges I think the install docs will be in a really good place (I don't have any more PRs planned for now).

Next up I'm going to take a look at some of the star specific stuff and might have some new docs then.

matildepark:
I disagree on what looks better but would take a trade off where you mark the OS lines as comments, so they aren’t given bash syntax highlighting

matildepark (approved):

PR #472, opened 2020-09-09
Edit providers copy

mhartl:
Redux of Providers copyedits.


mhartl:
Is there a way to auto-ignore `content/docs` changes in future?

matildepark:
Depends how you're committing your changes. You should always be precise with what's in the commit.

baudtack (approved):

PR #473, opened 2020-09-12
 Keep guide up to date with other docs

zalberico:
* Urbit install guide now uses parent urbit directory
* Digital Ocean changed plans from standard to basic and dropped the $15 3GB plan to 2GB of memory.
* DO boxes now use Ubuntu 20.04 by default
* Gandi changed their UI again for no reason

### DIFF
--- a/content/using/operations/hosting.md
+++ b/content/using/operations/hosting.md
@@ -18,8 +18,7 @@ This guide uses Digital Ocean as the cloud provider, but others can be used.
  - Create an account on [Digital Ocean][Digital Ocean].
  - Create a droplet with the following settings:
  - **Image**: Ubuntu 20.04 x64
- - **Plan**: Basic
- - **Size**: $20 (4GB/2CPU)
+ - **Plan**: 4GB RAM
  - **Add block storage**: Skip
  - **Datacenter Region**: Choose the region closest to you.
  - **VPC Network**: No VPC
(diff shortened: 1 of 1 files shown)


zalberico:
@matildepark , @baudtack , @botter-nidnul for SA

content/using/operations/hosting.md
```diff
@@ -17,12 +17,9 @@ This guide uses Digital Ocean as the cloud provider, but others can be used.
...
  - Create a droplet with the following settings:
- - **Image**: Ubuntu 18.04 x64
- - **Plan**: Standard
- - **Size**: 
-   - **Minimum**: $10 (2GB/1CPU) 
-   - **Recommended**: $15 (3GB/1CPU) 
-   - **Bonus**: $20 (4GB/2CPU)
+ - **Image**: Ubuntu 20.04 x64
+ - **Plan**: Basic
+ - **Size**: $20 (4GB/2CPU)
```
matildepark: The exact plan doesn't matter. Just specify how much RAM you'll want.

matildepark (commented):
All other changes are fine.

content/using/operations/hosting.md
zalberico: I've been suggesting 4GB because the swap stuff of 2GB and below seems like a pain, is that what I should do here? or say at least 3GB?

2GB with a comment about swap?

zalberico: I just set it to 4GB

zalberico:
Good to merge?

matildepark (approved):

zalberico:
Thanks!

PR #482, opened 2020-09-29
index.html: gently tweaking top layout

matildepark:
cc @galenwp @urcades 

Screen Shot 2020-09-29 at 6 34 56 PM

### DIFF
--- a/templates/index.html
+++ b/templates/index.html
@@ -20,14 +20,7 @@
 {% endblock description %}
 {% block content %}
 
-<section class="full lh-copy c2-11-md c3-11-lg mt4 f6">
-  <strong>Urbit</strong>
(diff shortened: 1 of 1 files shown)


urcades:
I like this tweak! It settles my mind a bit, since the nav jump was really the only " " no ideal " " aspect to the last design that I had the biggest misgiving about. There's obviously still a visual jump here, but it seems more purposeful for some reason.

The only 'above the fold' CTA we'd be missing is the "Blog" nav link, which is simply located in the recent posts div. I think this is a fine tradeoff for a total site that feels more coherent.

matildepark:
If you compare deploy preview against live site, it frames the pitch really nicely and just looks less busy, well paced. I think it makes a really positive diff.

galenwp:
yeah this is fine by me

PR #489, opened 2020-10-13
add documentation for s3 setup

jalehman:
Adds docs on how to set up S3 bucket integration both from a Dojo and Landscape.

### DIFF
--- a/content/using/operations/using-your-ship.md
+++ b/content/using/operations/using-your-ship.md
@@ -578,15 +578,15 @@ Our S3 support lets you store credentials for your S3 object storage buckets on 
 
 To enable S3 support on your ship you'll need to first set up a bucket, and then add its credentials to Landscape using *either* Landscape or the Dojo.
 
-### 1. Bucket Setup
+### Bucket Setup
 
 If you’re using AWS, it will have to support signature v2, not v4. We've found DigitalOcean's Spaces to work well also. The bucket has to be publicly readable, allow CORS from `*` origins, allow GET and PUT methods, and allow `*` headers. Your provider should have details on setting access permissions and CORS.
 
-### 2. Add your credentials: Landscape
+### Add your credentials: Landscape
 
 Navigate to your `https://<your-ship-url>/~profile/settings`, or click your sigil in the upper-right corner of the window. Fill out the **S3 Credentials** and **S3 Buckets** sections and you're all set!
 
-### 2. Add your credentials: Dojo
+### Add your credentials: Dojo
 
 Once you have gotten your bucket setup, poke the `s3-store` on your ship with your details. You can do this in Dojo or in web Dojo.
 
--- a/content/using/operations/using-your-ship.md
+++ b/content/using/operations/using-your-ship.md
@@ -576,23 +576,27 @@ Amazon S3, which stands for Amazon Simple Storage Service, is a technology that 
 
 Our S3 support lets you store credentials for your S3 object storage buckets on your Urbit ship; once you have done so, you will get additional functionality for uploading your own media within Chat and Groups (for avatars).
 
-### Bucket Setup
+To enable S3 support on your ship you'll need to first set up a bucket, and then add its credentials to Landscape using *either* Landscape or the Dojo.
 
-If you’re using AWS, it will have to support signature v2, not v4. We've found DigitalOcean to work well also. The bucket has to be publicly readable; allow CORS from `*` origins, and allow `*` headers. Your provider should have details on setting access permissions and CORS.
+### 1. Bucket Setup
 
-### Add your credentials: Landscape
+If you’re using AWS, it will have to support signature v2, not v4. We've found DigitalOcean's Spaces to work well also. The bucket has to be publicly readable, allow CORS from `*` origins, allow GET and PUT methods, and allow `*` headers. Your provider should have details on setting access permissions and CORS.
+
+### 2. Add your credentials: Landscape
 
 Navigate to your `https://<your-ship-url>/~profile/settings`, or click your sigil in the upper-right corner of the window. Fill out the **S3 Credentials** and **S3 Buckets** sections and you're all set!
 
-### Add your credentials: Dojo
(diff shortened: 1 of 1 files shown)


content/using/operations/using-your-ship.md
```diff
@@ -570,6 +570,33 @@ Syntax `;unset notify`
...
 
+## S3 {#s3}
+
+Amazon S3, which stands for Amazon Simple Storage Service, is a technology that was introduced by Amazon Web Services for cloud storage that has since been extended more broadly into a [standard programming interface](https://en.wikipedia.org/wiki/Amazon_S3#S3_API_and_competing_services).
+
+Our S3 support lets you store credentials for your S3 object storage buckets on your Urbit ship; once you have done so, you will get additional functionality for uploading your own media within Chat and Groups (for avatars).
+
+### Bucket Setup
+
+If you’re using AWS, it will have to support signature v2, not v4. We've found DigitalOcean to work well also. The bucket has to be publicly readable; allow CORS from `*` origins, and allow `*` headers. Your provider should have details on setting access permissions and CORS.
```
tylershuster: Might say "DigitalOcean's 'Spaces'" — it took me a minute to find the right one

content/using/operations/using-your-ship.md
```diff
@@ -570,6 +570,33 @@ Syntax `;unset notify`
...
+
+### Add your credentials: Landscape
+
+Navigate to your `https://<your-ship-url>/~profile/settings`, or click your sigil in the upper-right corner of the window. Fill out the **S3 Credentials** and **S3 Buckets** sections and you're all set!
+
+### Add your credentials: Dojo
```
belisarius222, line 589: I think it should be clarified that you should use *either* landscape or dojo here; these are alternatives, and you don't have to do both.

content/using/operations/using-your-ship.md
tylershuster: Also something about the allowed methods — I think GET and PUT are required

content/using/operations/using-your-ship.md
```diff
@@ -570,6 +570,33 @@ Syntax `;unset notify`
...
+
+Once you have gotten your bucket setup, poke the `s3-store` on your ship with your details. You can do this in Dojo or in web Dojo.
+
+```
+:s3-store|set-endpoint 'endpoint.mys3provider.com’
```
belisarius222: What is the endpoint?  Do I need to "create an access point" for my bucket?

content/using/operations/using-your-ship.md
```diff
@@ -570,6 +570,33 @@ Syntax `;unset notify`
...
+:s3-store|set-access-key-id ‘MYACCESSKEYID'
+:s3-store|set-secret-access-key ‘MYSECRETACCESSKEY’
```
tylershuster: Change these from smart quotes to dumb ones

content/using/operations/using-your-ship.md
belisarius222: Or maybe this is just the domain ... I think it would be good to specify what this is for actual Amazon S3, not just the general case where it could be something else pretending to be S3.

content/using/operations/using-your-ship.md
```diff
@@ -570,6 +570,33 @@ Syntax `;unset notify`
 Do not notify when your ship name is mentioned.
 
 
+## S3 {#s3}
+
+Amazon S3, which stands for Amazon Simple Storage Service, is a technology that was introduced by Amazon Web Services for cloud storage that has since been extended more broadly into a [standard programming interface](https://en.wikipedia.org/wiki/Amazon_S3#S3_API_and_competing_services).
+
```
tylershuster, line 576: I might add something about a link to Minio like "If you want to stay completely self-hosted you can try minio but here there be dragons." I know it works but it's tricky and for the more technically inclined only.

content/using/operations/using-your-ship.md
matildepark: The endpoint is indeed the domain. S3 has diff signature versions per endpoint that should be coordinated.

belisarius222: Ok.  How do I set the signature version? Does that go into the domain somehow, is it a setting elsewhere in the AWS console, or both?

jalehman: When creating a new bucket I don't even see an option for setting signature version. Should we just remove this instruction @matildepark ?

content/using/operations/using-your-ship.md
jalehman: I'm totally unfamiliar with Minio, so I'd prefer to leave this out unless someone else wants to contribute the docs.

content/using/operations/using-your-ship.md
```diff
@@ -576,23 +576,27 @@ Amazon S3, which stands for Amazon Simple Storage Service, is a technology that
...
 
-### Add your credentials: Landscape
+If you’re using AWS, it will have to support signature v2, not v4. We've found DigitalOcean's Spaces to work well also. The bucket has to be publicly readable, allow CORS from `*` origins, allow GET and PUT methods, and allow `*` headers. Your provider should have details on setting access permissions and CORS.
+
+### 2. Add your credentials: Landscape
 
 Navigate to your `https://<your-ship-url>/~profile/settings`, or click your sigil in the upper-right corner of the window. Fill out the **S3 Credentials** and **S3 Buckets** sections and you're all set!
 
-### Add your credentials: Dojo
+### 2. Add your credentials: Dojo
```
tylershuster: should this be 2b or 3?

content/using/operations/using-your-ship.md
matildepark: https://docs.aws.amazon.com/general/latest/gr/s3.html

content/using/operations/using-your-ship.md
jalehman: No, the idea is that there's only one step 2—either adding credentials via the Dojo, *or* via Landscape.

PR #490, opened 2020-10-16
add geturbitid.com to install page

cxk280:
`geturbitid.com` was tested with positive feedback from several Tlon employees. At this point it also has several paying customers sourced elsewhere, none of whom have sent in any complaints. Spoke about this to @syzygyzer and he told me to go ahead and send the PR in.

### DIFF
--- a/content/using/install.md
+++ b/content/using/install.md
@@ -162,8 +162,9 @@ If you have an Urbit ID, you'll use the [Bridge](https://bridge.urbit.org) clien
 
 Tlon doesn't currently sell or distribute Urbit IDs, but there are a few ways to get your own:
 
-- Getting an invitation from a friend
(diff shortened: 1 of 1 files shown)


drbeefsupreme:
geturbitid is a different category of service from the others, since it provides hosting in addition to selling IDs. So I think you should put it on its own bullet point here.

We probably also want some kind of explanation of what hosting providers are somewhere, but I think that ought to be a separate PR that we can handle.

tylershuster:
urth.systems offers a similar service to geturbitid and should probably be listed in the same category.

matildepark:
I don't believe urth actually offers hosting.

tylershuster:
You're right

cxk280:
@syzygyzer I pushed up a change based on your comment. Should be ready to go now.

When you refer to the separate PR about what hosting is, does `we` mean Tlon employees?

drbeefsupreme:
@cxk280 Yes, I meant Tlon employees (though of course anyone is welcome to submit such a PR, I'm just guaranteeing that myself or @baudtack will do it soon if nobody else does)

PR #491, opened 2020-10-28
Add Note on Nginx Configuration

Sneagan:
The web client requires constant refreshes to receive new messages without these settings in an Nginx reverse proxy.


baudtack (commented):
I don't know anything about running urbit behind nginx. @vvisigoth any idea who might? Just want a second verification that this is correct.

Fang-:
Long-standing advice for fixing nginx sse shenanigans has been to follow [this StackOverflow advice](https://stackoverflow.com/a/13673298/1334324).

@Sneagan (sorry for the wait on this and) the PR here mentions only three of those. I'm assuming you're running a standard nginx setup, so does that make the explicit `Connection` header and http version optional, as far as landscape behavior is concerned? Just making sure we're not leaving anything important out.

Might also be worth mentioning, super briefly, that this goes into the `location` block in your nginx config.

baudtack:
Bump.

baudtack:
@vvisigoth Do you know who might have the nginx knowledge to review and verify that this pr is technically correct?

baudtack:
@brendanhay Do you happen to have any insight into nginx to tell if this pr is technically sound?

brendanhay:
@baudtack LGTM

PR #492, opened 2020-10-29
added OTA documentation

tylershuster:
@syzygyzer moved https://github.com/urbit/docs/pull/954 over here. I didn't think it was appropriate under *Landscape*, so see here

### DIFF
--- a/content/using/operations/using-your-ship.md
+++ b/content/using/operations/using-your-ship.md
@@ -104,7 +104,7 @@ Ctrl-y    Yank from kill buffer
 
 ### Updates
 
-By default, you will automatically receive updates ([OTAs](@/docs/glossary/ota-updates.md)) from your sponsor. To check your OTA source, run `|ota` in the [dojo]((@/docs/glossary/dojo.md).
+By default, you will automatically receive updates ([OTAs](@/docs/glossary/ota-updates.md)) from your sponsor. To check your OTA source, run `|ota` in the [dojo](@/docs/glossary/dojo.md).
 
 If for some reason (for example, if your sponsor is out of date), you can switch OTA sources by running `|ota ~otasrc %kids` in the dojo, where `~otasrc` is the ship from which you want to receive updates. It is a good idea to contact the source ship and ask permission to sync from them.
 
--- a/content/using/operations/using-your-ship.md
+++ b/content/using/operations/using-your-ship.md
@@ -102,6 +102,14 @@ Ctrl-u    Kill to beginning of line
 Ctrl-y    Yank from kill buffer
 ```
 
+### Updates
+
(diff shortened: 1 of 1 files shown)


drbeefsupreme:
By "under Landscape" I meant in its own section at the same level as Landscape, not as a part of that section. Above it works fine too. LGTM!

drbeefsupreme:
Premature, premature! From Oct 15 OTA:

* Added %take-this, %take-that, %meet-this, and %meet-that merge
stragies, as defined in pkg/arvo/gen/hood/merge/help.txt, which is
printed when |merge is run with no arguments.  Usually %take-that is
preferable to the old %that.

* Renamed %this and %that to %only-this and %only-that to emphasize
their destructive nature

* Thus, the new process for checking whether you're up-to-date is
  - check your %base-hash.  If it's up-to-date, you're good
  - else, check your %sour-hash.  If it's not up to date, then you're either
    sill downloading the update or not able to connect to your OTA source.
    Wait or switch your OTA source with |ota ~new-ship %kids
  - once your %sour-hash is up-to-date, if it's not equal to your %base-hash,
    run |merge %home (sein:title our now our) %kids, =gem %take-that
  - this will overwrite any changes made to files in the main distribution,
    but it won't overwrite any files you added, such as publish notebooks.
  - note any errors and check again if your %base-hash is equal to %sour-hash
  - now your %base-hash and %sour-

content/using/operations/using-your-ship.md
```diff
@@ -102,6 +102,14 @@ Ctrl-u    Kill to beginning of line
 Ctrl-y    Yank from kill buffer
 ```
 
+### Updates
+
+By default, you will automatically receive updates ([OTAs](@/docs/glossary/ota-updates.md)) from your sponsor. To check your OTA source, run `|ota` in the [dojo](@/docs/glossary/dojo.md).
+
+If for some reason (for example, if your sponsor is out of date), you can switch OTA sources by running `|ota ~otasrc %kids` in the dojo, where `~otasrc` is the ship from which you want to receive updates. It is a good idea to contact the source ship and ask permission to sync from them.
+
+If OTAs are not succeeding, or if you are on an version of Urbit before the `|ota` command was introduced, you can run `|merge %kids (sein:title our now our) %kids, =gem %that`. **Warning:** this will wipe out any documents in the publish app, as well as any custom code. It will leave chats, links, and group memberships intact.
```
matildepark, line 111: 1. Should be "Warning: This" for capitalisation
2. This isn't actually true anymore for Publish (as of last night), it uses Gall state like every other app now.

matildepark: Also yeah %take-that will remedy this issue as a whole anyway

tylershuster:
R'ed new P

PR #494, opened 2020-10-30
ota new syntax

tylershuster:
@syzygyzer @matildepark New syntax, updated warning to note.

### DIFF
--- a/content/using/operations/using-your-ship.md
+++ b/content/using/operations/using-your-ship.md
@@ -108,7 +108,7 @@ By default, you will automatically receive updates ([OTAs](@/docs/glossary/ota-u
 
 If for some reason (for example, if your sponsor is out of date), you can switch OTA sources by running `|ota ~otasrc %kids` in the dojo, where `~otasrc` is the ship from which you want to receive updates. It is a good idea to contact the source ship and ask permission to sync from them.
 
-If OTAs are not succeeding, or if you are on an version of Urbit before the `|ota` command was introduced, you can run `|merge %kids (sein:title our now our) %kids, =gem %that`. **Warning:** this will wipe out any documents in the publish app, as well as any custom code. It will leave chats, links, and group memberships intact.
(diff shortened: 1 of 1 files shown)


content/using/operations/using-your-ship.md
```diff
@@ -108,7 +108,7 @@ By default, you will automatically receive updates ([OTAs](@/docs/glossary/ota-u
 
 If for some reason (for example, if your sponsor is out of date), you can switch OTA sources by running `|ota ~otasrc %kids` in the dojo, where `~otasrc` is the ship from which you want to receive updates. It is a good idea to contact the source ship and ask permission to sync from them.
 
-If OTAs are not succeeding, or if you are on an version of Urbit before the `|ota` command was introduced, you can run `|merge %kids (sein:title our now our) %kids, =gem %that`. **Warning:** this will wipe out any documents in the publish app, as well as any custom code. It will leave chats, links, and group memberships intact.
+If OTAs are not succeeding, or if you are on an version of Urbit before the `|ota` command was introduced, you can run `|merge %kids (sein:title our now our) %kids, =gem %take-that`. **Note:** This will wipe out any custom code that conflicts with code in the OTA. This should not be a problem unless you are developing core.
```
drbeefsupreme: Based on the OTA notes:

> this will overwrite any changes made to files in the main distribution,
> but it won't overwrite any files you added, such as publish notebooks.

it sounds like *any* changes to files in the distribution will be overwritten, not just ones that conflict with code in the OTA. Maybe this is what you meant, but if so its not clear to me. I'd be happy with just replacing "custom code that conflicts with code in the OTA" with "custom changes made to the base distribution".

Also this may just be me not being up to snuff on programming lingo yet, but "developing core" doesn't have a well-defined meaning in my head (though tbh neither does "base distribution"), and its made worse by the fact that core is a well-defined term in Hoon. Not sure what to replace it with, but I'd be fine with dropping it entirely since it seems covered by the previous sentence and instead including a blurb about how it won't affect your notebooks, chats, etc.

tylershuster: Ok, I think this is better.

PR #495, opened 2020-11-03
Clarify mentions of specified Ames port

lycurgus:
Also add the port to the firewall commands


drbeefsupreme:
LGTM but hosting docs is more @baudtack 's purview

drbeefsupreme (approved):

content/using/operations/hosting.md
```diff
@@ -83,11 +83,15 @@ Continuing to follow the DO docs we're going to configure the ufw firewall.
    ```
    $ sudo ufw app list
    ```
- - Next we'll configure ufw to allow connections via ssh and to allow Urbit to use the standard web port when the firewall is enabled.
+ - Next we'll configure ufw to allow connections via ssh and to allow Urbit to use the standard web port when the firewall is enabled,
+ as well as opening a port that we'll later use for Ames, Urbit's p2p protocol.
```
baudtack: I can't really put my finger on why I don't like referring to Ames as urbit's p2p protocol.

baudtack (changes requested):

content/using/operations/hosting.md
lycurgus: How about "opening a port that we'll later specify for your urbit to use to communicate directly with other ships"?

baudtack (approved):

PR #496, opened 2020-11-09
star and galaxy hosting correction

drbeefsupreme:
Removed some incorrect copy about our plans for hosting software. Linked to our
documentation on hosting. Clarified that you really ought to run your star if
you distribute planets, but it is not a requirement and the planets under it won't be doomed.

Based on a request in the UC docs channel

----

#

### DIFF
--- a/content/using/operations/stars-and-galaxies.md
+++ b/content/using/operations/stars-and-galaxies.md
@@ -8,9 +8,11 @@ template = "page_indiced.html"
 
 To boot your galaxy or star, follow our [installation instructions](@/using/install.md).
 
-### Hosting your star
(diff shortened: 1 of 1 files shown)


content/using/operations/stars-and-galaxies.md
```diff
@@ -8,13 +8,14 @@ template = "page_indiced.html"
 
 To boot your galaxy or star, follow our [installation instructions](@/using/install.md).
 
-### Hosting your star
+### Hosting your star
```
baudtack: Kill the space

baudtack (changes requested):

PR #498, opened 2020-11-12
interplanetary commerce: amends to avoid duplicate copy

matildepark:
(See email at feedback@)


jalehman:
Thanks for doing this @matildepark — turns out that this blog post went up prematurely and isn't supposed to be there anyway, so I took it down here: https://github.com/urbit/urbit.org/pull/499

Going to close this out.

PR #500, opened 2020-11-16
Explain unresponsive prompt

afknapping:
During install I was at first able to execute `(add 2 2)`, then I the `<<http>>` spinnter came up more often until the prompt became unresponsive to any keyboard input. Incidentally, that behaviour behaviour was explained at https://discord.com/channels/586329492260257812/596041386063691786/776134117590695936 and https://discord.com/channels/586329492260257812/596041386063691786/776134249098903642 and was helpful to both the person asking and me reading along.

### DIFF
--- a/content/using/install.md
+++ b/content/using/install.md
@@ -232,6 +232,8 @@ Or, if you'd prefer to copy your key in, you can run:
 
 Either command will create a directory called `sampel-palnet/` and begin building your ship. It may take a few minutes.
 
+Dojo might be unresponsive for some time (minutes) during first boot, for example you might see a `<<http>>` or `<<behn>>` spinner and keyboard is unresponsive.
+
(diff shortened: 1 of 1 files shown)


content/using/install.md
```diff
@@ -232,6 +232,8 @@ Or, if you'd prefer to copy your key in, you can run:
 
 Either command will create a directory called `sampel-palnet/` and begin building your ship. It may take a few minutes.
 
+Dojo might be unresponsive for some time (minutes) during first boot, for example you might see a `<<http>>` or `<<behn>>` spinner and keyboard is unresponsive.
```
baudtack, line 235: But what is it doing? If you add a short "an indication of activity in eyre, urbit's http vane" and something like that for behn that would be great.

baudtack (changes requested):

content/using/install.md
baudtack: Bumping this pr.

afknapping: Apologies, have way too little knowledge yet to explain what is happening, so will close.

PR #509, opened 2020-11-26
breach: eagerly recommend flagday tool

Fang-:
Also took the liberty of making the rest of these urls relative, for cleanliness.

image

We might consider reordering the sections from this page to put the "what do I need to do" bits higher up on the page.


Fang-:
Oops, looks like these were Chesterton's absolute links.

jalehman (approved):

matildepark:
Relative links need to end in .md (like mine was in step 4).

PR #510, opened 2020-11-27
OTA command fix

drbeefsupreme:
The suggested command to run if OTAs were not succeeding had you merging into
your %kids desk instead of %home. This fixes that.

----

#


drbeefsupreme:
I'm getting a bit cheekier with merging my own PR's lately if they're minor enough. Someone smack me if I'm getting out of line!

PR #515, opened 2020-11-30
adds subdomain to hosting acme instructions

drbeefsupreme:
The `:acme` command should be correct per @joemfb, just getting an OK on copy.

----

#


jalehman (approved):
Much clearer.

PR #517, opened 2020-12-01
security and continuity post

galenwp:


galenwp:
okay, this is ready. who wants to push the big green button?

PR #521, opened 2020-12-07
turn off your ship for flagday

drbeefsupreme:
Should probably also mention this in the flagday script.

----

#


drbeefsupreme:
This might seem obvious but someone in UC asked it

philipcmonk (approved):
Sure.  `flagday` will kill a running process but this is nicer.

PR #522, opened 2020-12-07
OS1 has been released!

drbeefsupreme:
Updating the introduction page. This is the right file, right? It's not a
markdown file and I don't know how the website actually works, but grepping for
the text in it only turned up this page. LMK if its something else.

----

#


galenwp:
This looks right to me. 

We should probably update the naming to call it 'Landscape', but that might require changes all over the place.

matildepark:
It's indeed the right place. I'm hitting the button

PR #529, opened 2020-12-15
voting: update instructions for latest poll

Fang-:
We'll want to replace these with more generic and broader instructions at some point, but for now this is fine.

### DIFF
--- a/content/voting-instructions.md
+++ b/content/voting-instructions.md
@@ -7,9 +7,9 @@ hidetitle = "true"
 
 ## Arvo Stability Vote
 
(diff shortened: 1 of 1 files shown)


vvisigoth (approved):
Looks extremely gud

content/voting-instructions.md
```diff
@@ -7,9 +7,9 @@ hidetitle = "true"
 
 ## Arvo Stability Vote
```
joemfb, line 8: We should probably change the title too.

PR #530, opened 2020-12-15
Remove references to v0.10.8 & fix binary paths

ritfed-tabryg:
The install & ops guides mentioned invoking urbit from a version-dependent directory that is both out-of-date and unnecessary now that there is a stable URL for the latest binary.

Some people on the urbit-discord stumbled on this, either by downloading an old binary or not having the directory listed in the examples.


drbeefsupreme:
Thanks!

PR #536, opened 2020-12-18
install: Clarify wget/curl and WSL2 instructions

fischerando:
The install instructions were recently changed to use `wget` rather than `curl`, but `wget` isn't available by default on MacOS, so I restored the `curl` instructions for those lines. I *believe* that `curl -JLO` replicates all the behaviors of `wget --content-disposition` but I welcome correction.

I also updated one line which still referenced `curl` on Ubuntu, and added a link to `~sitful-hatred`'s [awesome WSL2 instructions](https://subject.network/posts/urbit-wsl2/) in the Windows section.

This is my first pull request, so if I've violated any style guidelines, thank you for your patience!


matildepark (approved):
Works as expected.

PR #537, opened 2020-12-18
clarify what a breach is

galenwp:
it's true that 'breach' sounds like a security incident. updated the language here to clarify for people who have never seen urbit before.

### DIFF
--- a/content/breach.md
+++ b/content/breach.md
@@ -1,5 +1,5 @@
 +++
-title = "~2020.12 Network Breach"
+title = "~2020.12 Network Reset"
 template = "page_indiced.html"
 +++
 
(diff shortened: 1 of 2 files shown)


matildepark:
But no one calls it a reset? They call it a network-wide breach. You could just say "breached continuity" to be specific.

baudtack (changes requested):
Agree with @matildepark continuity breach would be better.

galenwp:
Everyone is going to think that anything with the word 'breach' in it is a security breach. Just went through this with someone else this morning.

Blockchain-based systems often do 'hard forks' or 'resets'. Feel free to suggest something that doesn't involve the word 'breach' on the frontpage, where someone may not already know inside-urbit terminology.

vvisigoth:
We should merge this unless we get a better idea from @matildepark @baudtack that doessn't involve the term "breach."

You have until EOD to submit ideas.

matildepark:
I don't have any, just hit the button

PR #545, opened 2021-01-09
Add OTA setup instructions for comets

ritfed-tabryg:
Comets don't come with OTAs enabled by default, probably for a few reasons:
- UX for comets wasn't a focus until recently, when getting a planet became a significant barrier to onboarding
- Many comets are ephemeral and only used for testing, where the boot pill normally suffices and updating to the very latest version is wasteful
- Stars hosting comets don't want to pay for "freeloaders" and want to do as little as possible for non-Sybil-resistant ships to prevent DDoS attacks

However, doing support for comets in urbit-help & on Discord, this seems to be one of the top stumbling blocks to getting set up. Some people have had issues joining certain channels in urbit-community with the urbit-v1.0 pill hash (which is probably another ticket in itself...) which have magically gone away once they were up to date.

Here I explain how to enable OTAs for comets (there was no existing section for that since planets do it by default) via their sponsor, if only so I can direct people with this issue to this section.

### DIFF
--- a/content/using/install.md
+++ b/content/using/install.md
@@ -127,6 +127,27 @@ Good, your Dojo is working correctly. Now let's run our first useful command in 
 
 The `>=` output means that a command was successful. Now you can see your ship's files in its Unix directory.
 
+### Updating your comet
+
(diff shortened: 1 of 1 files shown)


jalehman (approved):
This is a great addition—thanks for adding!

ritfed-tabryg:
Any blockers to merging this? It's less urgent now that the pill has been updated & isn't incompatible with a bunch of chats in urbit-community.

PR #546, opened 2021-01-14
use less discouraging language, add hosting path

jalehman:
The existing introduction to setting up Urbit is somewhat discouraging (and wrong), and there's no reference to Tlon's hosting service or clarification about what hosting even is.

This solution is a stopgap to a more comprehensive treatment of the "Getting on Urbit" journey given the high volume of new interest that we have right now.

### DIFF
--- a/content/using/install.md
+++ b/content/using/install.md
@@ -6,9 +6,9 @@ template = "page_indiced.html"
 aliases = ["/docs/getting-started/"]
 +++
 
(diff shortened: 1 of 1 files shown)


galenwp:
reading through the commit, this looks good to me. haven't rendered it, but i think the language is nice.

galenwp (approved):

PR #547, opened 2021-01-14
make language less harsh

jalehman:
Lots of feedback has indicated that this page makes it seem like we've had a serious security incident, which is a fundamentally absurd proposition.

### DIFF
--- a/content/reset.md
+++ b/content/reset.md
@@ -5,15 +5,15 @@ template = "page_indiced.html"
 
 ## TL;DR
 
-*Last updated ~2020.12.9*
+*Last updated ~2020.1.14*
 
-A network reset happened on ~2020.12.8 — you may need to take action. We call an event like this a 'breach' since we breach the continuity of Urbit's network protocol. This isn't a security incident.
+A network reset happened on ~2020.12.8 — you may need to take action.
 
-Unlike past breaches, this one comes with a tool to export and reimport all of your data so that you don't lose anything; we call it `flagday`. Details on its usage can be found below.
+Unlike past resets, this one comes with a tool to export and reimport all of your data so that you don't lose anything; we call it `flagday`. Details on its usage can be found below.
 
 *If you are self-hosting your ship, following these instructions is mandatory in order to communicate with the rest of the network.*
 
-## When was the last network breach?
+## When was the last network reset?
 
 On Tuesday, ~2020.12.8.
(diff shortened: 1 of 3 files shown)


content/reset.md
```diff
@@ -5,46 +5,46 @@ template = "page_indiced.html"
 
 ## TL;DR
 
-*Last updated ~2020.12.9*
+*Last updated ~2020.1.14*
```
matildepark: Year

jalehman: Doh

ohAitch:
"Continuity breach" is long-established terminology; one is in fact a serious incident, if not a security one, even in intentional cases. Given tlon has recently come out with a moderately strong commitment to avoid network-wide breaches going forward, now is not the time to try to tone down the language for optics.

I do agree that calling it a "network breach" is perhaps much, but "reset" is understating it: urbit as a network is designed to reach a steady-state that does not have flag days, any continuity breach is a violation of the core premises outlined in the 100-year computer post. The terminology of a reset generally invokes a reset *button* and/or the system discarding transient state on power loss. This is not an appropriate metaphor.

tylershuster:
I agree that the word "reset" causes problems because it sounds like there is a button somewhere that can be pressed that levels the network. 

However, I also agree that "breach," however well-established semantically in a certain context, also causes problems. To the layman - which we are starting to attract - a breach is like a breach in a ship's hull, or a breach in security. Given the current environment, this suggests highly troublesome things to someone considering the project. Placing an alarming banner at the top of the home page insinuates parler-level problems - and first impressions can be everything for someone searching for a new online home. 

Perhaps now that daily users have all worked out the kinks of the breach, a banner of a more pleasant color like blue would be appropriate with positive text such as "The development team has deployed a new era on the network. If you have previously but not recently run Urbit, you may need to upgrade. Read more here"

ohAitch:
Fortunately, Tlon is [optimistic that this [was] the last time we ever do it](https://urbit.org/blog/security-and-continuity/), so alarming users can be avoided by the simple expedient of not doing any more breaches. Nor expecting users to "work out the kinks".

If that statement is materially false, and dozens of breaches are planned over the next decade, it would perhaps entirely behoove the users to be alarmed, and rather inappropriate to try to be positive about breaking the trust in the architectural vision of the network.

PR #549, opened 2021-01-20
Update value-of-address-space-pt2.md

eriknewton:
Minor changes for clarity


crrncllns (approved):
1. Change these dashes to em dashes without spaces (—): we – as builders – 
2. Insert comma between 'hand' and 'would': on the other hand would
3. Remove extra space after 'that': it's worth noting that
4. Insert comma btw 'being' and 'buying': For the time being buying and selling of Urbit IDs
5. Remove extra space after 'the': counter to the  intent
6. Insert comma after 'land': land as it were

PR #550, opened 2021-01-20
Update value-of-address-space-pt3.md

eriknewton:


crrncllns (approved):
1. Remove comma after 'year' and remove extra space after 'now': year, and now  unlock
2. Insert hyphen: urbit.org owned
3. Insert 'we do' before 'stars and planets': more about that ownership than stars and planets

Issue #285, opened 2019-11-15
Make headings clickable for #anchored URLs


Fang-:
A lot of our docs pages are fairly long and have multiple sections. Often, I want to link someone to a specific part of the page. We support `/some/page#anchor` behavior for most of our headings, but having to append that manually (and sometimes, figuring out the right anchor text to use) it a bit burdensome. It would be great if we could click headers, or a 🖇  next to them, to navigate to its `#anchor` on the page. Similar to how Github allows you to do in readme's and other `.md` files.

matildepark:
This is an easy addition but I’m curious about the best design for it.

Fang-:
I'd say making them invisibly clickable is fine. Expected behavior here (as per GH's implementation) is that there's a visual indicator on-hover, and just the clickable-cursor might suffice.

(But I'm not Design Team, which should probably still weigh in.)

ynx0:
I was missing this feature as well. I think that readthedocs.io does this well, where they show a some sort of symbol next to the title on hover, along with some alt-text saying something like "Permalink to this headline". I have seen a pilcrow and a chain (to symbolize a link) being used.

* [Example with chain](https://scapy.readthedocs.io/en/latest/build_dissect.html#layers)
![chain](https://user-images.githubusercontent.com/9275716/105128201-8fe7a900-5ab0-11eb-9f88-e4127d2d5cc5.png)

* [Example with pilcrow](https://jupyter.readthedocs.io/en/latest/)
![pilcrow](https://user-images.githubusercontent.com/9275716/105128045-2c5d7b80-5ab0-11eb-8d55-9f4bb3f1b59c.png)

tylershuster:
https://www.getzola.org/documentation/content/linking/#anchor-insertion

ynx0:
I decided to try my hand at it and created a draft pr (#551). Please let me know what you think.

ynx0:
I think #551 resolves this.

Fang-:
Doesn't look like #551 affected the getting started section of the docs, but otherwise, hurray!

ynx0:
Oh yeah, that's cause I didn't know which parts of the docs to enable them so I just did a couple of them that made sense at the time... Enabling/disabling can be done with the `anchors.py` script.

PR #551, opened 2021-01-20
Add anchor links

ynx0:
This is a draft pr for adding anchor links as discussed in #285.  

I have created the `anchor-link.html` template and added styling for it in `_main.scss`. 
Then, one can add `insert_anchor_links = "right"` to any section's front matter and expect it to work ([docs](https://www.getzola.org/documentation/content/linking/#anchor-insertion)). I tried it out with `urbit.org/contents/docs` and it seemed to work nicely, but I didn't know where it would actually be appropriate to put it so I haven't made that change yet. Also, since it is a submodule I don't know exactly how I would update that alongside the urbit.org repo. 

Since I am new to contributing I don't exactly know the conventions for the repo, so please let me know if you need changes made before merging.

### DIFF
--- a/sass/_main.scss
+++ b/sass/_main.scss
@@ -650,3 +650,22 @@ hr {
 nav::-webkit-scrollbar {
     display: none;
 }
+
+
+// custom anchor styling
+a.zola-anchor {
+    visibility: hidden;
+    padding: 0 0.1rem;
+    &:hover {
+	background-color: black;
+	color: white;
+    }
+}
+
+h1, h2, h3, h4, h5, h6 {
+    &:hover > a.zola-anchor {
+        visibility: visible;
+        border-bottom-style: none;
+    }
+}
+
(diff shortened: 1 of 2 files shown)


matildepark:
I would put both the option to set this in front matter; and make it on by default in some templates. Then the docs templates can get the anchors without needing to update every page in the Urbit/docs repo.

ynx0:
Hmm, I know what you mean but I don't know how to make it enabled by default on all children article on docs. From the [zola documentation](https://www.getzola.org/documentation/content/linking/#anchor-insertion), it says that the "`insert_anchor_links` property is set at the section level". We'd want to avoid manually setting this in every header like you said, but even after looking for a bit, it's not obvious to me how we could accomplish that.

matildepark:
Then we'll want a related PR on the section templates on urbit/docs:

https://github.com/urbit/docs/blob/master/_index.md etc etc.

ynx0:
Relevant PR filed at https://github.com/urbit/docs/pull/1014. I think it makes sense for this one to be merged first. I'll change the submodule back to the original docs and then everything should work.

matildepark (changes requested):
Creates some general hijinks I had to fix by forcibly updating the submodule. Could you just rebase this and remove the submodule changes entirely? It's not really relevant to the PR.

ynx0:
Ok, wow figuring out that I needed to force push after rebasing took forever but I it should work now... Let me know if it still needs some prodding.

matildepark (approved):
lgtm

PR #556, opened 2021-01-28
More copy improvements

jalehman:
There are a few things rolled into this:

* FAQ: I've moved the **Objections to Urbit** header to the bottom and renamed it to **Miscellaneous**. Calling out "objections" is a good way to make an otherwise unaware person aware that such objections exist in the first place. There's no need for this.
* Breach Banner: We're nearly two months past this and it looks kind of scary. If anyone's opposed, I'll put it back.
* Install + Setup: I've renamed this link to **Join the Network**, re-captioned it with copy that doesn't mention the CLI, and moved it to the top of the list. This is our only real Call to Action, so it should be prominent.

### DIFF
--- a/content/faq.md
+++ b/content/faq.md
@@ -31,72 +31,6 @@ We encourage outside contributors to become a part of the project. The best way 
 
 [Tlon](https://tlon.io), a San Francisco-based company, is the primary developer of Urbit, along with [various independent developers](https://github.com/urbit/urbit/graphs/contributors) and [urbit.live](https://urbit.live).
 
-## Objections to Urbit
-
-### Who is Curtis? {#who-is-curtis}
-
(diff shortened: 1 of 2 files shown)


drbeefsupreme (changes requested):
I definitely agree with moving the objections further down the page and renaming to Miscellaneous.

I think the "Why are there not more planets" and "Why is it currently expensive to acquire a planet" (which actually has a typo - it says "is is" instead of "is it") should not be moved so far though - perhaps put them under the Urbid ID section.

Same for "Can I host my Urbit with someone else" - maybe under Urbit ID as well, but I'm less sure about that. We definitely want people to know hosting is available, and the FAQ is rather long.

matildepark (approved):

galenwp (approved):
works for me!

jalehman:
> Same for "Can I host my Urbit with someone else" - maybe under Urbit ID as well, but I'm less sure about that. We definitely want people to know hosting is available, and the FAQ is rather long.

Doesn't really seem like an Urbit ID related question. I think this question is adequately answered in other places, like the install guide, so I'm less concerned with this being front and center.

PR #559, opened 2021-02-03
Remove diff information from hosting page

samblarson:
It looks like some diff information got committed on the hosting page. This removes it.


matildepark (approved):
Silly merge conflicts.

PR #561, opened 2021-02-09
bifurcated install guide

tylershuster:
In this PR I have taken the current language in the install guide and split it into two pages: "Getting Started" and "Setting up your Planet" — I think this gets people on board faster, while still making the requisite information available. I think there's still work to be done in the planet guide, but I haven't touched much of anything in there. I have updated some of the language in the "Getting Started" guide to be less technical, more results-oriented. I've also added a little tab switcher to the OS chooser, which I think ameliorates some of the earlier problems about which OS should come first (though I do still think the Mac tab should be first).

I don't know what's going on with the `content/docs` directory; let me know if you know how to fix that.

### DIFF
--- a/content/using/install.md
+++ b/content/using/install.md
@@ -118,7 +118,7 @@ There are five kinds of Urbit IDs but to get started you only need to know about
 
 ***
 
-**Planets** are scarce (this prevents spamming, among other things) and usually require a price to acquire. As plans for Urbit come to fruition, most people will use these.
+**Planets** are scarce (this prevents spamming, among other things) and usually require a price to acquire. This is the recommended way to run Urbit, but it's a bit more involved.
 
 A planet name looks like `~sampel-palnet`.
 
@@ -207,6 +207,8 @@ The dojo is Urbit's command line. You can use it to control your ship, or to exe
 
 ### Purchase a Permanent Identity {#boot-your-planet}
 
-You can continue using this comet indefinitely. There are currently few differences between using a comet-level identity and a planet-level one. However, some groups will not allow comets entry in order to maintain a certain level of quality, and changes may be made in the future that further devalue comets. They will always, however, be able to access the basic functions of the network. A comet also comes with a long and fairly unmemorable name whereas a planet has a short name and a "sigil" (avatar) associated with it that makes it more identifiable on the network. You may notice all this within the first few minutes of using Urbit.
+You can continue using this comet indefinitely. There are currently few differences between using a comet-level identity and a planet-level one. However, some groups will not allow comets entry in order to maintain a certain level of quality, and changes may be made in the future that further devalue comets. They will always, however, be able to access the basic functions of the network.
+
+A comet also comes with a long and fairly unmemorable name whereas a planet has a short name and a "sigil" (avatar) associated with it that makes it more identifiable on the network. You may notice all this within the first few minutes of using Urbit.
 
 [To read instructions on how to purchase and use a planet, visit this page](/using/planet).
--- a/content/using/install.md
+++ b/content/using/install.md
@@ -143,7 +143,7 @@ To boot a comet, go into the command line and run the following command from the
 ~/urbit/urbit -c mycomet
 ```
 
-It may take a while to load the comet. This comes along with it being free. It won't take more than an hour, but will probably only take a few minutes. When it's done you'll some messages ending like this:
+It may take a while to load the comet (probably only take a few minutes, but it could take longer). This comes along with it being free. When it's done you'll some messages ending like this:
 
 ```
 ames: on localhost, UDP 31337.
@@ -178,7 +178,7 @@ At the moment, the most common way to use Urbit is through a web app named Lands
 
 ### Updating your comet {#updating}
 
-The Urbit binary comes with a somewhat recent release of the Urbit OS, but ships can update automatically ("over the air"), so new binaries aren't necessary every time it's updated. Planets have automatic updates enabled by default, but this is not the case for comets. Many comets are used only once and thrown away, so it would be wasteful to update every single comet as soon as it boots. If you plan to use your comet for more than a quick test, you'll probably want to ensure you're running the latest version of the OS.
+The Urbit application comes with a recent release of the Urbit OS, but automatic updates ("over the air") can be enabled so new binaries aren't necessary for each update. Planets have automatic updates enabled by default, but this is not the case for comets. Many comets are used only once and thrown away, so it would be wasteful to update every single comet as soon as it boots. If you plan to use your comet for more than a quick test, you'll probably want to ensure you're running the latest version of the OS.
 
 You can enable updates for your comet by typing `|ota (sein:title our now our) %kids` into Dojo and pressing Enter.
 
--- a/content/using/install.md
+++ b/content/using/install.md
@@ -197,7 +197,15 @@ use |ota %disable or |ota ~sponsor %kids to reset it
 >=
 ```
 
-### Purchase a Permanent Identity {#planet}
+### Read More about the Dojo {#the-dojo}
+
+The dojo is Urbit's command line. You can use it to control your ship, or to execute arbitrary code. Check out the links below for more information.
+
+- [Basic Operations](/using/operations/using-your-ship)
+- [Basic Hoon](/docs/tutorials/hoon/hoon-school/setup)
+- [Glossary Entry](/docs/glossary/dojo)
+
+### Purchase a Permanent Identity {#boot-your-planet}
 
 You can continue using this comet indefinitely. There are currently few differences between using a comet-level identity and a planet-level one. However, some groups will not allow comets entry in order to maintain a certain level of quality, and changes may be made in the future that further devalue comets. They will always, however, be able to access the basic functions of the network. A comet also comes with a long and fairly unmemorable name whereas a planet has a short name and a "sigil" (avatar) associated with it that makes it more identifiable on the network. You may notice all this within the first few minutes of using Urbit.
 
--- a/content/using/install.md
+++ b/content/using/install.md
@@ -1,75 +1,149 @@
 +++
-title = "Install + Setup"
+title = "Getting Started"
 weight = 1
 description = "How to install Urbit."
 template = "page_indiced.html"
 aliases = ["/docs/getting-started/"]
 +++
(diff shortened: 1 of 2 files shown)


tylershuster:
Looks like there's a broken link in docs. How do I update those when they're a separate repo?

matildepark:
There's several broken links in docs — all to sections of the install guide that were changed / removed here. You have to update docs stepwise to amend the relative link and then merge urbit.org/master into your PR branch again (pushing to master on docs prompts an auto-pull on this repo).

tylershuster:
It sounds like you're saying I have to have a PR accepted on docs before I can do this update but that doesn't sound right...

matildepark:
And yet it is!

tylershuster:
Is it possible to just add redirects from those URLs to the new locations?

matildepark:
You can remove the old relative links entirely. I don’t think you can make redirects for section headers specifically.

content/using/planet.md
```diff
@@ -3,7 +3,7 @@ title = "Setting up Your Planet"
 weight = 1
 description = "Getting started with a permanent Urbit identity."
 template = "page_indiced.html"
-aliases = ["/using/install.md#boot-your-planet", "/using/install.md#the-dojo"]
+aliases = ["@/using/install.md#boot-your-planet", "@/using/install.md#the-dojo"]
```
matildepark: Not how this works. The reason internal link format uses .md is that it scours the actual files and directories.

tylershuster: Oh, oops. I got a false positive that it worked on my machine. But I had changed the linking file

tylershuster:
Ok, at least this gets to a point where it's previewable

tylershuster:
@jalehman

content/using/install.md
```diff
@@ -1,246 +1,212 @@
...
-You'll want to install Urbit binary first.  Follow the instructions below to install the binary on Windows, MacOS, or Linux.
+There are currently two ways to run Urbit:
 
-### Windows
+If you use Mac or Linux and are not afraid of a few terminal commands, we recommend you run it yourself. That is the aim of this guide.
 
-> Please note that this method of installing Urbit is experimental, and we may not be able to assist you if you encounter issues related to WSL 2.
-
-Urbit cannot run on Windows itself, but there is a convenient way to run Linux using the [Windows Subsystem for Linux 2](https://docs.microsoft.com/en-us/windows/wsl/wsl2-install) on Windows 10. Install the Windows Subsystem for Linux 2 and open a Linux terminal in Windows, then follow the Linux installation instructions below. These instructions have been tested and verified for WSL 2 + Ubuntu 18.04 LTS, as demonstrated in `~sitful-hatred`'s step-by-step setup guide [here](https://subject.network/posts/urbit-wsl2/).
+If you run Windows or just want to give it a shot without worrying about the gritty bits, you may want to consider a [hosting provider](/using/planet#hosting-providers).
```
jalehman, line 19: Does this conflict with the fact that we have a path documented below to actually running Urbit on Windows?

content/using/install.md
```diff
@@ -1,246 +1,212 @@
...
+The **ID** is similar to a username and password on another site. The difference is that the public part (the username) is verified by thousands of people around the world while only you hold the key (the password) to that ID.
 
-### Hosting Providers {#hosting-providers}
+There are five kinds of Urbit IDs but to get started you only need to know about two: we call them **planets** and **comets**.
 
-Hosting providers will often sell you a planet and run it for you. This option is very simple but will probably cost a regular fee.
+***
 
-Urbit is designed to be portable. This means that if you sign up for hosting now but later want to leave your hosting provider and run your Urbit yourself, you should be able to work with them to obtain all of your data and boot your planet back up without losing anything.
+**Planets** are scarce (this prevents spamming, among other things) and usually require a price to acquire. As plans for Urbit come to fruition, most people will use these.
```
jalehman: > As plans for Urbit come to fruition, most people will use these.

"plans...come to fruition" is a little vague. 

Perhaps, instead, something like: "This is the recommended way to run Urbit, but it's a bit more involved."

content/using/install.md
```diff
@@ -1,246 +1,212 @@
...
-Open your keyfile (it is just a text file) and copy the key inside.  Paste it into the following command, with `sampel-palnet` replaced by the name of your planet.  **Do not include the tilde in your planet name.**
+To boot a comet, go into the command line and run the following command from the `urbit` directory you created during [Urbit installation](#installing-urbit):
 
 ```sh
-./urbit -w sampel-palnet -G rAnDoMkEy
+~/urbit/urbit -c mycomet
 ```
 
-Or you can copy the keyfile into the current directory and run:
+It may take a while to load the comet. This comes along with it being free. It won't take more than an hour, but will probably only take a few minutes. When it's done you'll some messages ending like this:
```
jalehman:
```suggestion
It may take a while to load the comet (probably only take a few minutes, but it could take longer). This comes along with it being free. When it's done you'll some messages ending like this:
```

content/using/install.md
```diff
@@ -1,246 +1,212 @@
...
-
-To start your planet up again, run the following from your `urbit` directory.
-
-```sh
-./urbit sampel-palnet
-```
+## Using Landscape
 
-Note that `sampel-palnet/` is the path of a folder, which we just created in your `urbit` directory. This folder is called your planet's **pier**, and it holds all of your data.
+Urbit is an entirely new computer, so by default it drops you into the "dojo," which is its terminal. Like your current computer, it can be used to do just about anything if you know the right commands. You only need to know one command for now.
```
jalehman, line 168: > You only need to know one command for now.

What's that command?

content/using/install.md
```diff
@@ -1,246 +1,212 @@
...
-Delete the keyfile with the command below:
+## Next Steps
 
-```sh
-rm ./my-planet.key
-```
+### Updating your comet {#updating}
 
-## Setup {#setup}
+The Urbit binary comes with a somewhat recent release of the Urbit OS, but ships can update automatically ("over the air"), so new binaries aren't necessary every time it's updated. Planets have automatic updates enabled by default, but this is not the case for comets. Many comets are used only once and thrown away, so it would be wasteful to update every single comet as soon as it boots. If you plan to use your comet for more than a quick test, you'll probably want to ensure you're running the latest version of the OS.
```
jalehman:
```suggestion
The Urbit application comes with a recent release of the Urbit OS, but automatic updates ("over the air") can be enabled so new binaries aren't necessary for each update. Planets have automatic updates enabled by default, but this is not the case for comets. Many comets are used only once and thrown away, so it would be wasteful to update every single comet as soon as it boots. If you plan to use your comet for more than a quick test, you'll probably want to ensure you're running the latest version of the OS.
```

Removed the use of the word "ship" since we haven't introduced that yet.

content/using/install.md
```diff
@@ -1,246 +1,212 @@
...
-From the Urbit Community group you can join a variety of chatrooms, notebooks, and collections.
+- [Basic Operations](/using/operations/using-your-ship)
+- [Basic Hoon](/docs/tutorials/hoon/hoon-school/setup)
+- [Glossary Entry](/docs/glossary/dojo)
 
-### Next Steps
+### Purchase a Permanent Identity {#boot-your-planet}
 
-Now that you’re up and running, take some time to deepen your understanding by checking out some of our other guides:
+You can continue using this comet indefinitely. There are currently few differences between using a comet-level identity and a planet-level one. However, some groups will not allow comets entry in order to maintain a certain level of quality, and changes may be made in the future that further devalue comets. They will always, however, be able to access the basic functions of the network. A comet also comes with a long and fairly unmemorable name whereas a planet has a short name and a "sigil" (avatar) associated with it that makes it more identifiable on the network. You may notice all this within the first few minutes of using Urbit.
```
jalehman: Big paragraph, could be broken up.

Would be nice to show a picture of a sigil here to drive the point home.

content/using/planet.md
```diff
@@ -0,0 +1,129 @@
++++
+title = "Setting up Your Planet"
+weight = 1
+description = "Getting started with a permanent Urbit identity."
+template = "page_indiced.html"
++++
+
+A planet is a permanent Urbit identity. It can never be taken from you, and its reputation will be tied to your own. If you're not yet familiar with the Urbit network, please follow the [Getting Started guide](/using/install), which will explain the basics and walk you through using a free identity. This guide explains the next steps.
```
jalehman: > It can never be taken from you, and its reputation will be tied to your own.

A comet can never be taken away from you either. I'm not sure that "its reputation will be tied to your own" is very clear to someone not already initiated.

content/using/planet.md
```diff
@@ -0,0 +1,129 @@
...
+Hosting providers will often sell you a planet and run it for you. This option is very simple but will probably cost a regular fee.
+
+Urbit is designed to be portable. This means that if you sign up for hosting now but later want to leave your hosting provider and run your Urbit yourself, you should be able to work with them to obtain all of your data and boot your planet back up without losing anything.
+
+Using hosting does mean that you're trusting your provider with your data, but so long as you have your planet, you'll always own your identity.
+
+Current hosting providers are:
+
+- [Tlon Corporation](https://tlon.io): Tlon is the creator of Landscape and is primarily responsible for the creation and maintenance of the Urbit project itself.
+- [Get an Urbit ID](https://www.geturbitid.com/): Get an Urbit ID was the first hosting provider and is run by a member of the Urbit community.
```
jalehman, line 31: Perhaps we should add https://www.escapepod.store/ in here too? cc @minderfolden for copy.

jalehman (changes requested):
I like that there's more meaningful context given to the reader about what it is that they're doing without getting too far into the weeds, but I think some small copy tweaks are needed.

content/using/install.md
tylershuster: Er, technically no. I just think that because most people who run Windows are probably not up for the longer install process, we should flag it. It could be something like "If you run Windows and don't want to deal with the slightly longer install process or just want to give it a shot..." but the language gets muddy.

content/using/install.md
tylershuster: +code

tylershuster:
These are good suggestions. I'll work on crafting solutions to the open-ended ones

jalehman (approved):

PR #563, opened 2021-02-23
add acme instructions

jalehman:
closes https://github.com/urbit/docs/issues/778

PR #565, opened 2021-02-24
Docs reorg

baudtack:


jalehman:
There are a bunch of broken relative links in this that prevent it from building—mind fixing @baudtack ?

baudtack:
On it.

matildepark:
This is crashing on the docs' repo's link to install.md. I also want to be cautious about changing that link — urbit.org/using/operating/install is a lil weird. We currently have an alias at urbit.org/install that redirects to /using/install that will also 404 here.

jalehman:
Yeah, I agree @matildepark. I think we'll want to keep the install/setup/quickstart guide (what's currently under /using/install.md) in its current place and with the same visual style. In other words, apart from the rest of the usage documentation.

drbeefsupreme:
I think this is missing changes to `static/glossary.js` I think @matildepark originally generated it automatically, so perhaps that may be done again.

baudtack:
@jalehman take another look? builds now in ci \o/

drbeefsupreme:
@baudtack `ship-troubleshooting.md` seems to be missing from `using/os`

jalehman (changes requested):
@baudtack The navigation structure isn't there. 

Going to `/using` should take me to an index that looks like `/docs`, but with a nav tree full of usage documentation (see the spec for reference). It looks like the folder structure is there, because I can go to `/using/id/guide-to-breaches`, which is the right URL—but the nav at left should be a menu that shows Urbit ID and Urbit OS headers. The nav appears to be unchanged from what it used to be: 

Screen Shot 2021-03-08 at 10 12 43 AM

At top left, it should look like this if you're in the `/using` subtree (operations-oriented):
Screen Shot 2021-03-08 at 10 21 48 AM

And like this when in the development docs:
Screen Shot 2021-03-08 at 10 21 28 AM

content/media.md
```diff
@@ -1,4 +0,0 @@
-+++
-title = "Media"
-template = "media/list.html"
-+++
```
jalehman: Why is this deleted?

matildepark:
You can add those links as catches right after this line: https://github.com/urbit/urbit.org/blob/master/templates/list.html#L23

Though you may want to hard code them, I'm unsure how to programmatically grab these links. You can see an if statement based on path here:

https://github.com/urbit/urbit.org/blob/master/templates/partials/pagination.html#L32

So `using/operations/_index.md` or whatever would be the relative path in question.

content/using/os/using-your-ship.md
```diff
@@ -0,0 +1,327 @@
++++
+title = "Using Your Ship"
```
jalehman: Perhaps a better title for this would be "Basics" or "Getting Started"? As in, this is the logical starting point when wanting to dig in to learning more about using your ship. Thoughts?

content/using/os/using-your-ship.md
```diff
@@ -0,0 +1,327 @@
...
+
+Otherwise, `-dns-auto` works the same as `:dns|ip` does with stars and planets: if it's available or unavailable, Chat messages, and so on.
+
+### More information
+
+Configuring a ship's domain causes the `:acme` app to request an HTTPS certificate for that domain from LetsEncrypt. Note that LetsEncrypt also requires that the HTTP server be listening on port 80. If the certificate request fails, `:acme` will send a `:chat-cli` message with an explanation. Once the certificate is successfully retrieved, `:acme` will install it, causing the HTTP servers to restart. A secure server will be started on port 443 (the HTTPS default) if it's available. Otherwise, it will try 8443, and then increment to the next port until it can successfully bind one.
+
+The built-in logic for listening on port 80 is similar. Urbit tries to bind port 80; if it cannot, it tries 8080, then increments until it can bind a port. Port 80 is available to unprivileged process on recent version of macOS. Otherwise, the process needs to either be run as root, or be given special permission (CAP\_NET_BIND on Linux).
```
jalehman: This content was [updated in a recent PR](https://github.com/urbit/urbit.org/pull/563/files). Let's make sure to not overwrite that.

content/using/planet.md
```diff
@@ -5,7 +5,7 @@ description = "Getting started with a permanent Urbit identity."
 template = "page_indiced.html"
 +++
 
-A planet is a permanent Urbit identity. Like all Urbit identities, they can never be taken away from you. Unlike comets which are free, planets are designed for use well into the future. If you're not yet familiar with the Urbit network, please follow the [Getting Started guide](/using/install), which will explain the basics and walk you through using a free identity. This guide explains the next steps.
+A planet is a permanent Urbit identity. Like all Urbit identities, they can never be taken away from you. Unlike comets which are free, planets are designed for use well into the future. If you're not yet familiar with the Urbit network, please follow the [Getting Started guide](@/using/install.md), which will explain the bafasdfsics and walk you through using a free identity. This guide explains the next steps.
```
jalehman:
```suggestion
A planet is a permanent Urbit identity. Like all Urbit identities, they can never be taken away from you. Unlike comets which are free, planets are designed for use well into the future. If you're not yet familiar with the Urbit network, please follow the [Getting Started guide](@/using/install.md), which will explain the basics and walk you through using a free identity. This guide explains the next steps.
```

jalehman (changes requested):
Added a few other notes.

jalehman:
@drbeefsupreme I added your redirects (plus some of my own modifications) to this branch, which should make for cleaner merging than merging your redirects separately into urbit.org. Going to close out https://github.com/urbit/urbit.org/pull/567 as a result.

jalehman (approved):

PR #569, opened 2021-03-12
troubleshooting: corrections and adjustments

Fang-:
Was browsing the preview for docs reorg, and my eye fell on some content on the "ship troubleshooting" page.

Pretty concerned about how often/easily this page says "you dun goofed, do a breach". If it's not user error, you shouldn't need to breach. At the very least, an issue needs to be reported.

Changes below are correction to inaccuracies, removing/changing outdated information, and toning down the breach-happiness a tiny bit. Happy to discuss.

(Targeting docs-reorg branch to avoid needing to do conflict resolution.)

### DIFF
--- a/content/using/os/ship-troubleshooting.md
+++ b/content/using/os/ship-troubleshooting.md
@@ -22,11 +22,11 @@ If your issue is not covered here, please check out our [Troubleshooting Knowled
 
 An ounce of prevention is worth a pound of cure, so let's first go over some best practices to keep your ship in working order.
 
-### Only boot your ship once
+### Only initialize your ship once
 
-Once your ship is booted, you should never boot it a second time. If you do boot it anew without special measures, you will experience trouble communicating with ships on the network you had talked to before.
+Once your ship has been initialized (with the `-w` variation of the `./urbit` command), you should never do so a second time. If you do initialize it anew without special measures, you will experience trouble communicating with ships on the network you had talked to before.
 
-If you accidentally booted a second time, or want to intentionally start fresh, you need to perform a [personal breach](#personal-breach), which is explained in the next section.
+If you accidentally did this a second time, or want to intentionally start fresh, you need to perform a [personal breach](#personal-breach), which is explained in the next section.
 
 ### Do not delete your pier
 
--- a/content/using/os/ship-troubleshooting.md
+++ b/content/using/os/ship-troubleshooting.md
@@ -22,19 +22,19 @@ If your issue is not covered here, please check out our [Troubleshooting Knowled
 
 An ounce of prevention is worth a pound of cure, so let's first go over some best practices to keep your ship in working order.
 
-### Only boot with your keyfile once
+### Only boot your ship once
 
-Once your ship is booted with your keyfile, you should never use that same keyfile again. If you do boot with the same keyfile twice, any other ship on the network that your ship has communicated with will not be able to talk to your ship.
+Once your ship is booted, you should never boot it a second time. If you do boot it anew without special measures, you will experience trouble communicating with ships on the network you had talked to before.
 
-If you accidentally booted with the same keyfile twice, the only remedy is performing a [personal breach](#personal-breach), which is explained in the next section.
+If you accidentally booted a second time, or want to intentionally start fresh, you need to perform a [personal breach](#personal-breach), which is explained in the next section.
 
 ### Do not delete your pier
 
-Urbit is stateful, meaning that it needs to hold onto all your data. If you delete your pier and start your ship again, you won't be able to talk to any ship you've talked to before. The only solution to this is performing a [personal breach](#personal-breach)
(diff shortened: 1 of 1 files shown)


content/using/os/ship-troubleshooting.md
```diff
@@ -22,19 +22,19 @@ If your issue is not covered here, please check out our [Troubleshooting Knowled
 
 An ounce of prevention is worth a pound of cure, so let's first go over some best practices to keep your ship in working order.
 
-### Only boot with your keyfile once
+### Only boot your ship once
 
-Once your ship is booted with your keyfile, you should never use that same keyfile again. If you do boot with the same keyfile twice, any other ship on the network that your ship has communicated with will not be able to talk to your ship.
+Once your ship is booted, you should never boot it a second time. If you do boot it anew without special measures, you will experience trouble communicating with ships on the network you had talked to before.
```
jalehman: What's "booting"? Is starting it again after you've shut it down booting?

That's obviously not what's meant here, but I could see someone being confused by that.

content/using/os/ship-troubleshooting.md
```diff
@@ -197,9 +193,13 @@ This means that your ship ran out of memory.
 
 2. Restart your ship. If you don't crash again, everything may be fine.
 
-3. Try the procedure outlined [here](https://github.com/urbit/support/wiki/ship-exits-with-a-bail:-meme-error).
+3. If you **do** crash again, try running the following after your ship has shut down:
+   `./urbit-worker meld your-ship` (Replacing `your-ship` with the name/directory of your ship.)
+   This will attempt to compact the memory of your ship. Note that this may use large amounts of memory on the machine you are running it on, and will be very slow if the machine has little memory available.
```
jalehman, line 198: 🙏

jalehman (commented):
I think the language around "never boot twice" could be confusing—what is "booting"? I think of booting as "turning it on", but you clearly mean "starting it anew with a keyfile."

matildepark:
You only boot once. Any other run of the Urbit binary is resuming a suspended ship.

jalehman:
> You only boot once. Any other run of the Urbit binary is resuming a suspended ship.

I understand that, but do you think it's clear to someone reading this for the first time?

tylershuster:
I only understand this as of yesterday, watching the old "what is urbit" video after actually having used it. I certainly thing it's an interesting concept to explain but it needs to be made explicit up front.

matildepark:
> > You only boot once. Any other run of the Urbit binary is resuming a suspended ship.
> 
> I understand that, but do you think it's clear to someone reading this for the first time?

Do you think it's clear why we call them gates? If you want people to know why running -w twice breaks everything, you should explain what it does.

content/using/os/ship-troubleshooting.md
baudtack: Hard agree. What we need is to fundamentally change what we call this operation. Boot is bad because everyone is confused by it. We have to stop using it. It may be the "correct" term but it brings too much baggage. For the same reason we call "functions" gates, we need to change this. @drbeefsupreme suggested last year we change it to "dawn" There was some technical objection to this but I think it works really well. That will have to be a secondary pr though as it will need to be consistently changed. 

Please revert this particular change.

baudtack:
> > > You only boot once. Any other run of the Urbit binary is resuming a suspended ship.
> > 
> > 
> > I understand that, but do you think it's clear to someone reading this for the first time?
> 
> Do you think it's clear why we call them gates? If you want people to know why running -w twice breaks everything, you should explain what it does.

The difference here is that gates, as a term, do not bring in an existing misunderstanding of what they are. Here people are used to booting and rebooting their computers. When you say "you can not boot an urbit twice", the immediate thought is  "uh... so I can't turn my computer off?" This confusion is a large source of support issues. As above, we intend to change this terminology but that will be done at a later time.

content/using/os/ship-troubleshooting.md
matildepark: Well I mean the old copy reads like you need to cycle networking keys and get a new keyfile every time you wanna resume your ship

baudtack (changes requested):
Everything looks good other than the boot coment.

content/using/os/ship-troubleshooting.md
drbeefsupreme: I agree with @jalehman and @baudtack that saying "never boot a second time" is confusing to new people. But also I agree with @matildepark that it shouldn't sound like you need a new keyfile every time. So I don't think it should be reverted to the old commit, but should take into account both of these concerns. And we'll change the language to dawn soon.

Fang-:
I went with "initialize" for now, and added explicit mention of this meaning "that command with the `-w` in it". Please opine.

baudtack:
> I went with "initialize" for now, and added explicit mention of this meaning "that command with the `-w` in it". Please opine.

More than four letters. F minus minus. :-p That works fine actually.

baudtack (approved):

PR #570, opened 2021-03-13
using-your-ship: use more recent, correct command to `Join the Urbit Community from Dojo`

botter-nidnul:


jalehman:
@botter-nidnul Hey ~botter—we just did a big overhaul of the docs. Mind resubmitting this to put the change at this location? https://urbit.org/using/os/messaging/

PR #572, opened 2021-03-19
simplify filesystem manual, remove wrong info

drbeefsupreme:
The Clay manual, which is in the `/using` section, was more technical than
`using.md` in the technical documentation. New users don't need to know what a
`beak` or `beam` is - just describing them is sufficient. So I've simplified the
terminology so that the only new term is `desk`.

Several commands were also out of date, wrong, or removed since it was first
written. They have been vanquished.

I was a little aggressive on the simplification, so whoever reviews this might
want me to re-add one or two things. But I'm pretty firm on how unnecessary it
is for every user to know what a `beak` is.

I might re-add some of these things to the technical documentation side, like
`beak` and `beam`.

### DIFF
--- a/content/using/os/filesystem.md
+++ b/content/using/os/filesystem.md
@@ -19,8 +19,9 @@ This quick-start guide will walk you through some common commands. Follow along
 using your Dojo. When you get a `>=` message after entering a command, this means
 that the command was successful.
 
-A `desk` is something like an independently revision-controlled branch of your
-urbit's file-system. Your urbit's system files live in the `%home` `desk`.
+A [`desk`](@/docs/glossary/desk.md) is something like an independently
+revision-controlled branch of your urbit's file-system. Your urbit's system
+files live in the `%home` `desk`.
 
 It's important to note that whenever you want to sync changes from your Unix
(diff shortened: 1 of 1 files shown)


content/using/os/filesystem.md
```diff
@@ -10,132 +10,139 @@ referentially transparent namespace. An easy way to think about it is like typed
...
+For more information on Clay, see the [Overview](@/docs/arvo/clay/clay.md), and
+additional usage information at [Using Clay](@/docs/arvo/clay/using.md). 
+
 ### Quickstart
 
 This quick-start guide will walk you through some common commands. Follow along
 using your Dojo. When you get a `>=` message after entering a command, this means
 that the command was successful.
 
+A `desk` is something like an independently revision-controlled branch of your
```
jalehman: Is there anything that we can link to about a [`desk`]? E.g. a glossary entry? If not, disregard.

jalehman (approved):
This is great. Once it builds, good to merge.

drbeefsupreme:
Looks like the problem with the build was actually from the docs submodule. Apparently just merging master, creating no changes, fixed it. But I'm guessing if I had merged it despite the broken build it would have been fine.

PR #577, opened 2021-03-26
Fixed dead links in the FAQ

simjnd:
The FAQ had some incorrectly formatted links which made them point to unexisting resources on the Urbit website.

This patch fixes these issues, and also points the answer about installation to the correct resource (here: https://urbit.org/getting-started/, **not** here: https://urbit.org/getting-started/planet/)


drbeefsupreme:
Thanks!

PR #581, opened 2021-04-02
Add links between Urbit address space posts

mhartl:
This PR adds links between the Urbit address space posts. Currently, only the final post has such links (and even then only in the main text). In particular, with this PR readers who land on the first post now have an easy way to find the second and third posts in the series.


drbeefsupreme:
Thanks!

drbeefsupreme:
The build is failing but I don't see why... the links look fine to me.

matildepark:
It commits docs submodule.

drbeefsupreme:
Ah right, the same reason another PR I merged earlier today failed.

mhartl:
I tried to make a commit without the documents submodule but relevant changes appeared to be present there. Is there a reference somewhere on what that submodule is and how to handle it?

matildepark:
The submodule is a local copy of urbit/docs. It gets auto-updated here when a commit hits master over there. You shouldn't commit the submodule because this repo handles it for you. 

You should instead `cd content/docs && git checkout master && git pull` to grab the submodule locally if you are trying to build the site on your computer and your local copy of the module is behind.

PR #584, opened 2021-04-02
fix #581

drbeefsupreme:
If anybody knows of a better way to clean up after merging a PR with an outdated docs submodule better than two reverts, branching off the original PR, updating the docs submodule, and then making another PR, let me know.


drbeefsupreme:
Now that I think about it, I'm guessing I just had to update the docs submodule directly, rather than this whole rigamarole.

matildepark:
More or less that would've worked yeah. In general I just don't let people commit the submodule

PR #586, opened 2021-04-02
adding -p ames_port to local install instructions pt 2

drbeefsupreme:
This is just #460 adapted to the new install directions. Not being terribly proficient with git, I wasn't sure how to handle the merge conflict when the page it was originally committed to having been split in two and most of the language being changed.

Notably, this only alters the instructions for planets - I left the instructions for comets the same. This is the only compromise I can think of that leaves the directions people new to Urbit are most likely to follow (the comet ones) as simple as possible, while addressing the fact that some people have issues with their Ames port. I fear that asking a new person to set up port forwarding to run a comet is just going to make a lot of people decide not to bother running Urbit at all. Most people don't know what that is or how to do it, and every hurdle we put up has attrition implications.

If it were possible to have some kind of data that tells us what percentage of people have Ames port issues, this would be an easier decision to make, but I don't see any easy way to go about that. @joemfb 's [suggestion](https://github.com/urbit/urbit.org/pull/460#issuecomment-736758545) on having your sponsor figure this out for you is the only reasonable suggestion I'm aware of.

### DIFF
--- a/content/getting-started/planet.md
+++ b/content/getting-started/planet.md
@@ -48,18 +48,26 @@ If you received an email invite to Urbit, the master ticket should be a .pdf fil
 
 ### Run the boot command {#the-dojo}
 
-Navigate to your `urbit` directory.
+For other ships on the network to communicate with your ship, the computer it's
+running on needs to be able to receive inbound UDP traffic. Therefore, it's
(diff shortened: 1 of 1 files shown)


jalehman (approved):
LGTM

PR #595, opened 2021-04-13
added youtube to events

tylershuster:
Adds optional youtube id to events and includes ids for past events. Shows youtube video in place of image on event page if present


jalehman (approved):
Gorgeous.

PR #596, opened 2021-04-13
new event, new template: bitcoin 2021

urcades:


jalehman:
This is so good @urcades

PR #601, opened 2021-04-16
site: extract email subscribe form to partial

tomholford:
This is a small refactor that consolidates multiple copies of the email subscription form into a single partial.

For context, I made this change while working on another branch and realized it would be an easy cherry-pick for `master`


jalehman:
Awesome. Thank you!

PR #603, opened 2021-04-20
config: fix rss feeds

### LINKED ISSUE #602: Fix RSS feed
Zola supports it, and the code is already in the `<head>` section:
```
<!-- START RSS -->
<link rel="alternative" type="application/rss+xml" title="RSS" href="/rss.xml" />
<!-- END RSS -->
```

But /rss.xml and /blog/rss.xml return 404.

matildepark:
As per [breaking changes in 0.11.0](https://github.com/getzola/zola/blob/master/CHANGELOG.md#breaking-1), RSS config variable and RSS template variable updated to reflect latest behaviour.

Fixes #602

### DIFF
--- a/config.toml
+++ b/config.toml
@@ -11,6 +11,7 @@ compile_sass = true
 # Whether to generate a RSS feed automatically
 generate_feed = true
 build_search_index = true
+feed_filename = "rss.xml"
 
 [markdown]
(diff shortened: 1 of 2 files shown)


tylershuster (approved):
LGTM, although should note that future events show up on the top of the feed until they pass.

PR #613, opened 2021-05-01
Fix broken URL in ship-troubleshooting.md

ghost:
The ship troubleshooting page (https://urbit.org/using/os/ship-troubleshooting/) linked to https://urbit.org/using/os/guide-to-breaches instead of https://urbit.org/using/id/guide-to-breaches/, which resulted in a 404 error.


tylershuster:
Thanks!

PR #620, opened 2021-05-06
blog post: naive rollups

drbeefsupreme:
Draft of a blog post on the current state of naive rollups. I attempted to write
for the least technical audience, aimed primarily on end user experience, with
links to the more technical stuff we've put out at the end.

The included image and title are provisionary. I couldn't help myself, IASIP is
my favorite show. Still needs a few more images.

Hope to get feedback from the infra/docs/L2 meetings next week, do some edits,
and publish late next week.

### DIFF
--- a/content/blog/rollups.md
+++ b/content/blog/rollups.md
@@ -12,16 +12,17 @@ Anybody who has used Ethereum smart contracts recently is aware that it has
 gotten _expensive_. There's all sorts of proposed solutions out there to make it
 affordable again-Ethereum 2.0 being the big one-but there's also a variety of
 complicated and clever mechanisms and proposals each with their own trade-offs
-that are collectively referred to as "layer 2 solutions". This is in contrast to
+that are collectively referred to as "layer 2 solutions" (see [further
+resources](#resources) for more information about these). This is in contrast to
 the first and original computational layer of Ethereum called "layer 1", where
 the brunt of the gas crisis is being felt.
 
 The standard personal vehicle for a human on Urbit is the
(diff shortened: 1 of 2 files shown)


drbeefsupreme:
Oh yeah, lingering questions about adoption between L1/L2 planets/stars. I just wrote that it doesn't matter, but I don't think this is settled. Since galaxies must be on L1, and stars can be on L1 with L2 spawn proxy, we obviously need to allow L1 sponsors with L2 sponsees. But I don't think L2 sponsors with L1 sponsees is decided yet? But in that case, its unclear what happens to L1 planets whose stars move to L2, so it seems like we need to allow everything.

basilesportif:
The only big question I feel like is missing here is the new security model: you should probably explicitly note that in L1, an Eth contract guarantees the validity of state transitions, while for L2 txs, each Urbit node is satisfying itself that it agrees with the state transitions. Even/especially non-technical people tend to be interested in what the security model is.

An alternate way of saying this is that the old model for the source of truth of a state transition was:
1. Run a transition
2. Eth enforces validity of the transition
3. an Urbit downloads the state from Eth
4. the Urbit implicitly gets a final veto on whether the state is valid, although that was never used

Now we are cutting out step 2, and beefing up step 4 to do actual validation.

basilesportif:
Looks great. One minor edit; you have:
> Your urbit gets a final decision on whether the new state is valid.

I would change "gets a" to "makes the", just because "gets" could be confused as meaning "receives".

drbeefsupreme:
Edits to this are happening on Google Docs now - DM me if you need a link

PR #626, opened 2021-05-20
hoon style redirect

drbeefsupreme:


matildepark:
In general I would advise more toward using `aliases` in the articles themselves if we can. See an example [here](https://www.getzola.org/documentation/content/page/).

PR #628, opened 2021-05-24
publish Reid's post

tirwyd-sarmes:


jalehman:
@tirwyd-sarmes Is the content here ready to merge? If so, I can do that—but first, please remove the commits to the docs submodule—those need to be made separately against the @urbit/docs repository. Thanks!

PR #629, opened 2021-05-26
Added groups to site

tylershuster:
Added all communities from the current index in draft state. We can add them as we clean them up or remove them if they're just not appropriate.

Currently Urbit Community and The Forge are published to start work on styling.

### DIFF
--- a/content/community/_index.md
+++ b/content/community/_index.md
@@ -1,5 +1,23 @@
 +++
 title = "Community"
-sort_by = "weight"
-template = "list.html"
+weight = 1
+description = "Urbit is for communities."
+page_template = "page_indiced.html"
+template = "community/index.html"
 +++
+
+# Urbit is for communities
+
+Whether your community is your family, your workplace, or another social circle, Urbit makes it easy to host a platform for discussion that can't be shut down or spied upon.
+
+## Groups
+
+Anyone can host a public group on Urbit. [Boot Urbit](/getting-started) or [click here](/community/groups) to browse public communities.
+
+## Meetups
+
+We meet up in real life to talk about Urbit and the future of computing. [Check out established meetups to find one near you.](/community/meetups/)
+
+## Governance
+
+[Learn more about the community that steers the direction of the project](/community/governance/).
--- a/content/community/groups/24-fps.md
+++ b/content/community/groups/24-fps.md
@@ -1,0 +1,15 @@
+
++++
+title = "24 FPS"
+weight = 118
+draft = true
+[extra]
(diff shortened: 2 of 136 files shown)


matildepark:
Dunno what the stance is on this project but it might be more sustainable to deploy it on a proxied static site after https://github.com/urbit/landscape/issues/61 and this project indeed may be a good enough push to do that work.

tylershuster:
I agree. However, @jalehman & I would like to start exposing parts of Urbit ASAP to drive adoption, and so this pass is basically `Dump Index state` > `Convert to Markdown` > `Publish the safe stuff` so that we can start to show "look people actually use this thing." We've talked about how to do this in a more automated manner, relying on the actual state of the network and our position at this point is that it's good to keep in mind but we shouldn't let it stop us.

Fang-:
Repeating from out-of-band discussion with @jalehman: what if there's people/groups who don't want their existence to be known outside of urbit? iirc index was always "an index _on urbit_"...

tylershuster:
Well, as it stands the groups would only be "public" in that they would be `.md` files in this repo. We'd only really publish groups on urbit.org after they had been vetted for both "not being insane" and "truly open to the public."

tylershuster:
image

tylershuster:
@urcades & I have discussed the design team's future plans for "group overview" visual design in light of the above screenshot and the knowledge that the tile metaphor is short-lived.

If this is ready for inclusion in the site before the tile metaphor is retired, I suggest that we push the design to the site to keep the metaphors consistent, and not to "spill the beans" on the new design ahead of time and risk confusing people, especially if the design changes significantly.

If the new design launches *before* this PR is ready, I think that updating the overview to reflect that design will be a relatively short task.

urcades:
Hello Everyone! 

I will make a note, that while I do rather like this section and love that it **spells out** what sort of communities live here, and that it's **even possible** to create communities, one of the chief issues with urbit.org as it stands is that it really isn't speaking to highly specific or concrete groups of people, and is quite muddy in terms of messaging when one takes a step back and assesses it as a whole — I don't believe fully that the addition of this section clears up the messaging, and adds yet another demographic (community builders/joiners) that the front page is addressing, further diluting what could be a very specific + strong front page.

This is to say, while I like this section as a hero, I do believe it'd either need to be a standalone hero image for it to make sense on the front page, or be designated to another page on the site, such as `urbit.org/community` or something.

As it stands, it competes with port to a light degree — happy to discuss/argue this one out with y'all, either on a call or in the context of this issue!

PR #630, opened 2021-05-27
Top matter refresh

urcades:
some small updates to the top matter of urbit.org

some of the links appear to be broken, I copied them over from the mp/port branch wholesale

PRing this so romina and I can perform minute tweaks to art


matildepark:
You're going to need this file — https://github.com/urbit/urbit.org/blob/mp/port/static/detect.js

and this line — https://github.com/urbit/urbit.org/blob/mp/port/templates/base.html#L63

urcades:
slamming that mf merge button, let's just do it

PR #632, opened 2021-05-27
static: fix OS autodetection

matildepark:


urcades (approved):
yes

matildepark:
That's actually introducing a grammatical error but whatever here we go

PR #634, opened 2021-05-27
fix link location for btc wallet

ixv:
Points the links on the bitcoin-wallet page to the right place


bioy-casares (commented):
looks good. thank you.

PR #635, opened 2021-06-01
blog: what's in a p

lycurgus:
I originally wrote this as an exploration of how the ID system works under the hood (particularly to help things like galaxy planets and star moons make sense). It was intended to just exist as a pastebin, or maybe a blog post on my own site somewhere - but I don't run a blog currently so it would have been fairly lonely. @urcades suggested it would be good as an urbit.org post and that raising a PR would be a simple way to get it in front of some eyes :)

Still some things to do:

- [ ] review the text for accuracy/comprehensibility/pacing
- [ ] change the header image? (the current one is Creative Commons but maybe something else would be preferred)
- [ ] host the final header image (currently it's on my own domain; not sure if I can include a file in this PR as well?)
- [ ] fix the date in the post header info
- [ ] anything else...?

### DIFF
--- a/content/blog/whats-in-a-p.md
+++ b/content/blog/whats-in-a-p.md
@@ -1,0 +1,180 @@
++++
+title = "What's In A @p?"
+date = 2021-06-01
+description = "An exploration of urbit's identity system"
+aliases = ["/posts/essays/whats-in-a-p", "/posts/whats-in-a-p"]
+[extra]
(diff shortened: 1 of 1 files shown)


urcades:
Added a few potential reviewers on this, but will give this a look myself soon!

tirwyd-sarmes:
I'm taking a look at this!

PR #636, opened 2021-06-01
20210601 grants update

mikeosborne:
Added 3rd party software distribution apprenticeship already completed by Raghu

PR #639, opened 2021-06-05
Grants reorg

mikeosborne:


mikeosborne:
It looks like this PR grabbed my commits from #636. Sorry, still learning git chops. Should I cancel this one until the others are merged?

jalehman:
It would probably make more sense to cancel the other one first. Let's just make sure to remove all of the duplicate grants, since now there's two of every one of them.

PR #653, opened 2021-06-18
updates "how secure is urbit?"

drbeefsupreme:
Somewhat recently, NIST pushed back their timeline on post-quantum cryptography.
Previously they had planned to release guidelines and recommendations sometime
next year. Now they anticipate releasing the results of their contest around the
end of this year, but won't release full recommendations until 2024.

I talked about this with @philipcmonk and have edited to FAQ to reflect this.
I've also added a few more details about other things, such as the fact that the
runtime is not audited and the event log is not encrypted at rest, and that Tlon
is looking to hire a cybersecurity expert. I also renamed the "Urbit HD Wallet"
section to "Urbit Security" since that section was not only about the HD wallet,
and moved the "How secure is Urbit?" question under that heading. This is
farther down the page. I don't think
its the most pressing question new users have at this point and so I don't think
it needs to be one of the first things on the FAQ.

### DIFF
--- a/content/faq.md
+++ b/content/faq.md
@@ -124,7 +124,7 @@ In late 2020, Urbit's [Ames](@/docs/glossary/ames.md) networking protocol was au
 
 The security of the runtime, [Vere](@/docs/vere/_index.md), has not yet been adequately assessed or systematically hardened.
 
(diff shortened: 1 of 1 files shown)


tirwyd-sarmes:
This looks great. 

Perhaps we can explicitly add more quotidian security information (rather than just referring to the docs)? Such as that messaging is AES encrypted. We can also specifically mention a couple areas under development like at-rest encryption.

philipcmonk (approved):

PR #658, opened 2021-06-23
removes reference to needing command line

drbeefsupreme:
Removed a line about needing the command line in the Understanding Urbit
section, since Port now exists. It's linked to on the previous line so I didn't
think it necessary to mention it explicitly. However, I did remove the link to
information about hosting providers. Since we're inviting someone to just check
things out, suggesting signing up for hosting seems like the wrong direction to
nudge rather than booting a comet via Port.


drbeefsupreme:
Not sure who should review this?

PR #659, opened 2021-06-23
removed extra header, fixed image link

drbeefsupreme:
The Understanding Urbit OS 1 -> OS N page had another header saying OS 1 ->
OS 3. Not sure which was supposed to remain, but I removed the 1 -> 3 one.

There was also a broken image link that I fixed.


drbeefsupreme:
I tagged @matildepark to review mostly because I figure you'd know what the header ought to be.

PR #662, opened 2021-06-23
json-jet.md: pin down consistency requirements

ohAitch:
(This may require some tweaks to the hoon implementation, with an eye for rejecting inputs hard to parse quickly)


jalehman:
Thanks @ohAitch !

PR #668, opened 2021-06-25
Docs routing and sidebar

g-a-v-i-n:
- swapped `docs/[...slug].js` for `docs/[[...slug.js]]` which lets the `docs.js` route and `/docs/*` route be handled by the same component
- minor modifications to support this in `lib.js`
- sidebar style pass
- moved sorting of layout tree out of frontend and into lib.js
- commented buildPageTree() mostly for my own benefit
- componentized docs sidebar and layout
- FAQ wip


lib/lib.js
```diff
@@ -144,35 +159,42 @@ export function formatDate(d) {
...
+          title: data.title,
+          base: page.name,
+          slug: page.name.replace(/.md$/, ""),
+          ...(ordering === "weight" && { weight: data?.weight ?? 0 }),
+          ...(ordering === "date" && {
+            date: data?.date ?? "2000-01-01T00:00:00.000Z",
+          }),
+        };
+      })
+      // Sort files according to weight. Weight can be a date or number. The choice between the two is set by the 'ordering' param.
```
matildepark, line 192: Actually not true, this will need a conditional on `type` to change sort algorithm

lib/lib.js
```diff
@@ -144,35 +159,42 @@ export function formatDate(d) {
 //     children: { [string]: PageTree },
 //   }
 
-export function buildPageTree(path, type = "", content = false) {
+export function buildPageTree(path, ordering = "", content = false) {
+  // get metadata from _index.md file in each folder. This is where page ordering (weight) is stored.
```
matildepark, line 163: Not true, it just gives us the parent directory’s metadata, weight is stored in each page’s metadata.

lib/lib.js
g-a-v-i-n: True, missed this

lib/lib.js
g-a-v-i-n: will update

PR #623, opened 2021-05-14
life and rift alterations

drbeefsupreme:
Addresses https://github.com/urbit/docs/issues/980

Related to https://github.com/urbit/docs/pull/1074 and it should be merged before this PR

This removes mention of life and rift from the guide to breaches. These are technical terms that I don't think the average user needs to know, and the using guide is supposed to be for the average user.

How to check a ship's life and rift is mentioned on the getting started guide, which I think is fine, but I've redirected the link in it from the guide to breaches to the new page on life and rift I've written.

PR #671, opened 2021-06-29
docs issue #623 again

baudtack:
it's not dns proxying, we have to stop calling it that.

https://github.com/urbit/docs/issues/623


jalehman (approved):
Looks great!

PR #723, opened 2021-07-28
urbit.org2: re-implementation in next.js

matildepark:
Live at https://urbit-org-gold.vercel.app.

- Migrates the entire urbit.org codebase from Zola to next.js. It requires a migration from Netlify to Vercel, and a change in our DNS accordingly, which will be completed after merge.
  - Iteration time when making changes to the site is now near-instantaneous, from a former 36s.
  - All pages load instantaneously, and can provide much more client-facing functionality, should we choose to incorporate more web applications on urbit.org itself.
- Redesigns the entire site with a friendlier look; also disposes of indigo-static in favour of Tailwind.
- Adds a navigation and filter pane for the new grants section.
- Supersedes https://github.com/urbit/docs, merging its entire Git history into the urbit.org repository. This was done because, while the project to maintain a separation of concerns was well-intentioned, it was not well-executed, and required a Git webhook to auto-commit docs changes as well as cross-repo coordination whenever any relative link was broken or changed on either repository. Given this is my fault over the last few years, I personally apologise for this. If and when docs becomes its own product, I imagine it will have to be written Urbit-upward from Sail and Udon, ported to the web format accordingly.

All features from the previous site installation should be retained or at parity in some sense — including all previous redirects, Twitter cards, RSS feeds, etc.


urcades:
Howdy, I just did a quick speedrun of the major leaves of the site (or tried to), including the top nav, footer content, outbound links, and I'm happy to report mostly everything seems good for release.

I'll note that I did catch a number of minor visual + state-representation issues that can be easily patched into the site once it's in production, as I venture that some of the issues I note might not apply to an actual live site.

Here's the list:

- I can’t tab through the links of the website — when I type tab at the top, it skips straight to the newsletter. (generally speaking, we should probably do an explicit accessibility pass as a distinct line of work after this release)
- On some youtube videos under /Events, I see "Playback on other websites has been disabled by the video owner" — we need to fix our youtube links for the site so they can render properly
- newsletter "sign up" button needs a state change to the highlight color (green) when an email has been properly written (in general, let's ensure form input handling is solid later)
- we need a new icon for "copy" button, for the urbit group reference
- on docs pages with a right-hand "header nav", the scrollbar for the page's main content should be butted up to the rightmost edge of the page, not “inside” or to the left of the right side bar nav
- the multi-selector for command line install ( macOS/linux/windows ) should be left-aligned with the following code (currently, the code appears indented)
- We'll need paged nav at the bottom of /getting started at some point
- little dot that indicates where in index you are should be made a little smaller
- when I press the next/previous buttons at the bottom of the scroll in the docs, I stay pinned to the bottom of the scroll — i should be bumped to the top of whatever article I’m about to read
- some of the links contained in the docs (i.e. within "system overview - hoon - further reading") are not green — it’s hard to tell they are links.
- In Hoon 2.5.2 , the next/previous buttons at the bottom are smushed completely together, rendering awkwardly
- we may need to check quote styling in the blog — I have the suspicion that quotes aren’t differentiated from normal text anymore
- When visited from the footer, Understanding Urbit’s (About) pages are not highlighted in the sidebar when you are viewing them, but the highlight reappears when you visit the next or previous section
- UU does not contain back/forward buttons
- It would be nice if the FAQ’s right-side header nav scrolled with you when you scrolled
- Maybe this is just the dev environment, but I seem to get broken pages if I use back/forward in the browser too much


Again, I'll note that these are all rather minor fixes we can easily make after release. The critical functionality of the site seems solid, and I'm excited to push this out to the world

urcades (approved):
Everything appears solid, and works as expected. There are some minor interface enhancements that have to be made for it to feel exceptional, but I think it's good to go.

matildepark:
> we may need to check quote styling in the blog — I have the suspicion that quotes aren’t differentiated from normal text anymore

Indeed it isn't, would you like to just add a border-left on those? Or do you have a mockup?

Can tackle some of these then merge.

matildepark:
Hold onto your hats.

PR #728, opened 2021-07-28
Update jael docs

tinnus-napbus:
here's the jael docs again.

I've fixed the links and headers to work with the new framework thingy. I've also removed any contents sections since they're redundant with the new sidebar contents thingy.

I've addressed all the issues raised by @Fang- in the original PR https://github.com/urbit/docs/pull/1106 except for:

> I don't think including the source comment here (and with other tasks) adds a lot. Consider taking them out.

cos I think it's useful to have a very succinct summary of the purpose of the task so you know what it does at a glance. Also they're not the source comment in all cases, I just used that in most cases because it was a good short summary. Maybe it could go somewhere else in the section, like above the type definition code? But I think they should be in there somewhere

@drbeefsupreme


content/docs/arvo/jael/jael.md
```diff
@@ -0,0 +1,25 @@
++++
+title = "Overview"
+weight = 1
+template = "doc.html"
++++
+
+Jael is the networking key infrastructure vane.
+
+The main purpose of Jael is keeping track of [Azimuth](/docs/azimuth/azimuth) related information. For each ship, this consists of its public keys, its key revision number (or `life`), its continuity breach number (or `rift`) (see [Life and Rift](/docs/azimuth/life-and-rift), and who the sponsor of the ship is. It also handles the local ship's private keys, keeps track of galaxy domain prefixes, and performs some boot-related tasks.
```
drbeefsupreme:
```suggestion
The main purpose of Jael is keeping track of [Azimuth](/docs/azimuth/azimuth) related information. For each ship, this consists of its public keys, its key revision number (or `life`), its continuity breach number (or `rift`) (see [Life and Rift](/docs/azimuth/life-and-rift)), and who the sponsor of the ship is. It also handles the local ship's private keys, keeps track of galaxy domain prefixes, and performs some boot-related tasks.
```

drbeefsupreme (changes requested):

content/docs/arvo/jael/jael.md
drbeefsupreme: I'm with @tinnus-napbus on the source comments. I think they're fine where they are, and ought to be somewhere, and that still seems like the best spot to me.

PR #739, opened 2021-07-28
Miscellaneous quality of life improvements to Grants

jalehman:


styles/markdown.module.css
```diff
@@ -95,12 +95,6 @@
   @apply py-2 ml-6 text-base md:text-lg lg:text-xl;
 }
 
-.markdown li a,
-.markdown h3 a,
-.markdown h2 a {
-  @apply text-green;
-}
-
```
matildepark: ???

jalehman: where did that that come from? definitely didn't edit this directly—is it generated and not supposed to be committed?

g-a-v-i-n:
Screen Shot 2021-07-28 at 1 36 30 PM

This works but is still visually confusing. @urcades was going to do a refactor of this whole UI I think. Correct me if I'm wrong, Ed

styles/markdown.module.css
matildepark: You should retain that css to keep all the links green inside posts — currently inside headers or lists links just have no styling

urcades:
@g-a-v-i-n yeah, absolutely — will get some suggestions floated into here so we can implement asap

styles/markdown.module.css
jalehman: Whoops. Added back.

jalehman:
> @g-a-v-i-n yeah, absolutely — will get some suggestions floated into here so we can implement asap

I'm going to just release this, and we can update the design afterwards.

PR #741, opened 2021-07-28
rss: fix link building

arthyn:
`path.join` was causing mishaps with link protocols, and really just for file paths.


matildepark (approved):
I thought I was so smart

PR #746, opened 2021-07-29
docs, contentarea: insert toc inside contentarea

matildepark:
Places the table of contents inside the content area so the scroll bar extends full page and the header on docs-style pages is full width.


urcades (approved):
This is a good change for now, fixes a few observed issues noticed by people in the community — we'll continue to tweak the right-hand sidebar for increased legibility, after this is merged.

Issue #735, opened 2021-07-28
Can't use arrow keys to select items in search results


ynx0:
What it says on the tin

g-a-v-i-n:
Resolved in #756

PR #756, opened 2021-08-02
font size, search refactor

### LINKED ISSUE #751: Opening a new docs page should scroll to the top of the screen
The way it works currently is kind of weird. If you scroll part of the way down a page and then open another page, you'll end up somewhere in the middle of that page. You should always start at the top.

g-a-v-i-n:
- Resolves #751 and #735
- Refactor search UI using Downshift
- Support keyboard search result browsing + selection
- Reduce global scaling (makes site visually smaller)
- Create two type scales - a larger one for blog-like content and a smaller one for documentation.


jalehman:
The experience of reading the blog on this PR and on the current (new) site is one of *density*. The old blog felt much more natural to read, so it would be good if the font size/column width/line height could be refactored to something approximating the old blog.

PR #764, opened 2021-08-06
update stdlib cork code, fix example

drbeefsupreme:
The code was out of date and the example no longer worked (`+mo` no longer
exists, and using `my` instead results in a `mull-grow` error)


drbeefsupreme:
Forgot this also updates the corl code

PR #769, opened 2021-08-09
add marks section to clay docs

tinnus-napbus:
- add overview
- add writing marks guide
- add using marks guide
- replace marks section of clay architecture doc with a link to the new
  marks section
- fix up links etc

### DIFF
--- a/content/docs/arvo/clay/marks/examples.md
+++ b/content/docs/arvo/clay/marks/examples.md
@@ -1,0 +1,248 @@
++++
+title = "Examples"
+weight = 4
+template = "doc.html"
++++
+
+These are the files used in the [Writing Marks](/docs/arvo/clay/marks/writing-marks) guide.
+
+## `/lib/csv/hoon`
+
+```hoon
+|%
+++  validate                                             ::  All rows same length?
+  |=  csv=(list (list @t))
+  ^-  ?
+  =/  result
+    %+  roll  csv
+    |=  [row=(list @t) =flag ref=(unit @ud)]
+    ?.  flag
+      [flag ref]
+    ?~  ref
+      [flag `(lent row)]
+    ?.  =(u.ref (lent row))
+      [%.n ref]
+    [flag ref]
+  flag.result
+++  en-csv                                              ::  csv -> atom (to mime)
+  |=  csv=(list (list @t))
+  |^  ^-  @t
+  (cat 3 (crip (join '\0a' (turn csv en-row))) '\0a')
+  ++  en-row                                            ::  encode row
+    |=  row=(list @t)
+    ^-  @t
+    (crip (join ',' (turn row en-field)))
+  ++  en-field                                          ::  encode field
+    |=  field=@t
+    ^-  @t
+    %+  rash
+      field
+    %+  cook
+      |=  =tape
+      ^-  @t
+      ?~  tape  ''
+      ?:  =('"' i.tape)
+        (cat 3 (crip tape) '"')
+      (crip tape)
+    ;~  pose
+      (full (star ;~(less (mask "\",\0a\0d") next)))
+      ;~  plug
+        (easy '"')
+        (star ;~(pose (cold '""' doq) next))
+      ==
+    ==
+  --
+++  de-csv                                               ::  atom -> csv (from mime)
+  |=  dat=@t
+  |^  ^-  (list (list @t))
+  =/  parsed  (parse dat)
+  ?.  (validate parsed)
+    ~|(%mixed-col-count !!)
+  parsed
+  ++  parse                                              ::  parse cord
+    |=  dat=@t
+    |^  ^-  (list (list @t))
+    %+  rash
+      dat
+    %+  most
+      ;~(pose (jest '\0d\0a') (mask "\0a\0d"))
+    (most com parse-field)
+    ++  parse-field                                      ::  parse field from cord
+      |^
+      %+  cook
+        |=(a=tape (crip a))
+      ;~(pose unenclosed enclosed)
+      ++  unenclosed                                     ::  unquoted field
+        ;~  sfix
+          (star ;~(less (mask "\",\0a\0d") next))
+          ;~  pose
+            ;~  plug
+              ;~(pose (jest '\0d\0a') (mask "\0a\0d"))
+              ;~(less next (easy ~))
+            ==
+            ;~(less doq (easy ~))
+          ==
+        ==
+      ++  enclosed                                       ::  quoted field
+        |^
+        ;~  sfix
+          (ifix [prefix suffix] content)
+          ;~  pose
+            ;~(simu (mask ",\0a\0d") (easy ~))
+            ;~  plug
+              ;~(pose (jest '\0d\0a') (mask "\0a\0d"))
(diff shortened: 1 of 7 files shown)


content/docs/arvo/clay/marks/writing-marks.md
```diff
@@ -0,0 +1,331 @@
...
+
+```
+foo,bar,baz
+blah,blah,blah
+1,2,3
+```
+
+There is a little complexity surrounding special characters in fields and line endings, but otherwise the only other rule is that all rows must have the same number of fields. You can refer to [RFC4180 on the IETF website](https://datatracker.ietf.org/doc/html/rfc4180) for more details.
+
+The simplest way to represent such a structure in urbit is as a `(list (list @t))` like:
```
drbeefsupreme:
```suggestion
The simplest way to represent such a structure in Hoon is as a `(list (list @t))` like:
```

content/docs/arvo/clay/marks/marks.md
```diff
@@ -0,0 +1,57 @@
++++
+title = "Overview"
+weight = 1
+template = "doc.html"
++++
+
+Clay is a typed filesystem, and we call these file types `mark`s. When talking about Hoon and Arvo we'll often talk of types like `@ud`, `(list @t)`, etc. A `mark` will specify such a type for its files, but it does more than just that - it also defines conversion routines to and from other `mark`s, as well as diff, patch, and merge routines.
```
drbeefsupreme, line 7: Somewhere on this page you should mention that marks are located in the `/mar` folder

content/docs/arvo/clay/marks/writing-marks.md
```diff
@@ -0,0 +1,331 @@
...
+
+So with the nature of the `%mime` `mark` hopefully now clear, the reason we want conversion methods to and from `%mime` in our `%csv` `mark` is so we can import CSV files from Unix and vice versa.
+
+Since a CSV file on Unix will just be a long string with ASCII or UTF-8 encoding, we can treat `q.q` in the `$mime` as a `cord`, and thus write a parser to convert it to a `(list (list @t))`. For this, I've created a `csv.hoon` library, which you can view in full on the [Examples](/docs/arvo/clay/marks/examples#libcsvhoon) page.
```
drbeefsupreme: "For this, I've created" First person sounds weird in documentation, maybe "We have a `csv.hoon` library...`?

content/docs/arvo/clay/marks/examples.md
```diff
@@ -0,0 +1,248 @@
...
+template = "doc.html"
++++
+
+These are the files used in the [Writing Marks](/docs/arvo/clay/marks/writing-marks) guide.
+
+## `/lib/csv/hoon`
+
+```hoon
+|%
+++  validate                                             ::  All rows same length?
```
drbeefsupreme, line 13: This is more of a side comment than a suggestion that you may have already considered, but it could potentially simplify some things to use `$|` to get a more restricted `(list (list @t))` type where all rows are the same length rather than calling `validate` in several locations. I haven't actually worked this through yet, and there might be pedagogical or stylistic reasons to use `(list (list @t))` rather than `$|  (list (list @t))  validate`, but perhaps it could be included as an exercise at the end if it actually makes sense to do so.

content/docs/arvo/clay/marks/examples.md
```diff
@@ -0,0 +1,248 @@
++++
+title = "Examples"
+weight = 4
+template = "doc.html"
++++
+
+These are the files used in the [Writing Marks](/docs/arvo/clay/marks/writing-marks) guide.
+
+## `/lib/csv/hoon`
```
drbeefsupreme, line 9: There's already a `/lib/csv/hoon` https://github.com/urbit/urbit/blob/master/pkg/arvo/lib/csv.hoon so this is probably going to confuse the reader without some clarification

content/docs/arvo/clay/marks/writing-marks.md
```diff
@@ -0,0 +1,331 @@
...
+ghi,jkl
+```
+
+So now our `%csv` `mark` lets us move data in and out of Urbit. In the next section, we'll look at the `+grad` arm in more detail.
+
+## `+grad`
+
+So far we've just delegated `+grad` functions to the `%noun` `mark`, but now we'll look at writing our own.
+
+For demonstrative purposes I've just poached the algorithms used in the `+grad` arm of the `%txt` `mark` and modified them to take our `(list (list @t))` type instead of a `wain`. It's not the most efficient algorithm for a CSV file but it'll do the job.
```
drbeefsupreme: Another use of first person

content/docs/arvo/clay/marks/using-marks.md
```diff
@@ -0,0 +1,240 @@
++++
+title = "Using Marks"
+weight = 3
+template = "doc.html"
++++
+
+In the last document, [Writing Marks](/docs/arvo/clay/marks/writing-marks), we walked through writing `mark` files and touched on how Clay handles them. They needn't just be left to background Vane processes though, you can also use them yourself in your code.
```
drbeefsupreme:
```suggestion
In the last document, [Writing Marks](/docs/arvo/clay/marks/writing-marks), we walked through writing `mark` files and touched on how Clay handles them. They needn't just be left to background vane processes though, you can also use them yourself in your code.
```

drbeefsupreme (changes requested):
Another great PR. Only some minor comments.

content/docs/arvo/clay/marks/examples.md
tinnus-napbus: oh true I wasn't familiar with `$|` tbh, I'll take a look and see if I can change it to using that

content/docs/arvo/clay/marks/examples.md
tinnus-napbus: dammit I never noticed that existed. I'll change the name of the library and add a clarifying note

tinnus-napbus:
@drbeefsupreme I have addressed the points you raised. Wrt the point about possibly using `$|`, I played around with it and I think it's simpler for demonstrative purposes to just use a `(list (list @t))`. Because the validation function in the `$|` is only called when you clam a noun as opposed to when you cast something, it would still require strategically clamming things similar to how I've put a `?>` in various places. So it's not all that much less complex while also having to define the type itself in the separate library and thus more complex in terms of how the walk-through proceeds.

Also master-morzod seemed to think it would be acceptable to do it this way

I have however added a note that you could use a `$|` instead and that we're doing it this way for the sake of demonstrative simplicity

drbeefsupreme:
LGTM! tagging @philipcmonk or @belisarius222 to check technical aspects

drbeefsupreme (approved):

tinnus-napbus:
ok `|crunch` introduced a new csv mark that conflicts with the example used in this doc. Since both the mark and lib now conflict with existing files, I've just changed them both to csv.hoon and added a note about how they conflict and that this isn't a problem on a fakezod

drbeefsupreme:
#1060 should be closed once this is merged

drbeefsupreme:
Still looking to get this checked @philipcmonk @belisarius222

Issue #780, opened 2021-08-13
remove use of `mo` from stdlib docs


drbeefsupreme:
I'm guessing `+mo` used to make `map`s?

example:
https://urbit.org/docs/hoon/reference/stdlib/2j

ohAitch:
What ever happened to it?

tinnus-napbus:
all mo's have been replaced now

PR #784, opened 2021-08-16
rewrite jar/jug examples to get rid of `mo`

drbeefsupreme:
Addresses #780

### DIFF
--- a/content/docs/hoon/reference/stdlib/2j.md
+++ b/content/docs/hoon/reference/stdlib/2j.md
@@ -18,8 +18,11 @@ A container arm for `++jar` operation arms. A `++jar` is a `++map` of
 `a` is a jar.
 
 ```
-    > ~(. ja (mo (limo a/"ho" b/"he" ~)))
-    <2.dgz [nlr([p={%a %b} q=""]) <414.fvk 101.jzo 1.ypj %164>]>
+> ~(. ja (my [a+1 b+2 ~]))
+< 2.ngd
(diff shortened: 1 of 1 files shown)


ohAitch:
Should these also be `ly` instead of `limo`?

drbeefsupreme:
What's the difference between them? I've been using them interchangeably.

```
++  le                                                ::  construct list
  |*  a=(list)
  ^+  =<  $
    |@  ++  $  ?:(*? ~ [i=(snag 0 a) t=$])
    --
  a

++  limo                                                ::  listify
  |*  a=*
  ^+  =<  $
    |@  ++  $  ?~(a ~ ?:(*? [i=-.a t=$] $(a +.a)))
    --
  a
```
I see that `+le` has a `list` as the sample, which seems circular? I also notice that they pretty print differently, and they are different nouns
```
> !=  (limo ~[1 2 3])
[8 [9 24.054 0 1.023] 9 2 10 [6 7 [0 3] 1 1 2 3 0] 0 2]
> !=  (ly ~[1 2 3])
[8 [9 3.062.778 0 1.023] 9 2 10 [6 7 [0 3] 1 1 2 3 0] 0 2]
```
but they're both described as producing a list from a null-terminated noun. Care to shed some light?

ohAitch:
Oh I just meant to be consistent about shortname constructors
Looks like `le` calls out to snag while `limo` does "unify the types of all the elements" explicitly? The former seems like it should break if you pass an unfaced list, if it does idk why it's in the constructor section.

drbeefsupreme:
Yeah I agree that using short name constructors for short name constructor examples is better stylistically.

PR #786, opened 2021-08-16
remove arvo overview ToC

drbeefsupreme:
no longer necessary because of the navigation on the right recently added


drbeefsupreme:
Whoops pushed to the wrong branch... ignore those commits

Issue #765, opened 2021-08-06
"using" docs are hard to find


drbeefsupreme:
There's a link to it at the bottom of the front page "Operator's Manual", and then another one on the page for setting up a planet: https://urbit.org/getting-started/planet

I think there should be a more prominent link on the front page, probably next to the Docs link at the top, and also one on the Getting Started page. Thoughts?

PR #787, opened 2021-08-16
make operator's manual easier to find

drbeefsupreme:
Addresses #765

Adds a link to the Operator's Manual (and Understanding Urbit) to the Getting
Started page. Should be uncontroversial.

Also adds a link to the Operator's Manual to the header on the front page. Definitely would
like feedback on this. Clutter on the front page should be avoided, and this
seems like the least useful link among the ones already in the header. So I
would be fine with removing it if anyone feels strongly.

Also changed the title of "Usage Docs" to "Operator's Manual", which is the name
we already use in a few places and sounds much less weird.

### DIFF
--- a/components/Header.js
+++ b/components/Header.js
@@ -77,6 +77,13 @@ export default function Header(props) {
           className="mr-5 type-ui"
           href="/blog"
         >
+          Operator's Manual
+        </ActiveLink>
+        <ActiveLink
+          currentPath={currentPath}
+          className="mr-5 type-ui"
+          href="/using"
+        >
           Blog
         </ActiveLink>
         <ActiveLink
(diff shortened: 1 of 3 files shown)


drbeefsupreme:
Added @marisarow  to review, is that appropriate or is someone else in charge of the look of the front page?

components/Header.js
```diff
@@ -72,6 +72,13 @@ export default function Header(props) {
         >
           Docs
         </ActiveLink>
+        <ActiveLink
+          currentPath={currentPath}
+          className="mr-5 type-ui"
+          href="/using"
+        >
+          Operator's Manual
```
jalehman:
```suggestion
          Manual
```

jalehman (approved):

PR #1089, opened 2021-08-27
add About and Grants to navbar

jalehman:
I think this is more useful content in the navbar—doesn't look terrible either. Thoughts @g-a-v-i-n ?

### DIFF
--- a/components/Header.js
+++ b/components/Header.js
@@ -68,9 +68,16 @@ export default function Header(props) {
         <ActiveLink
           currentPath={currentPath}
           className="mr-5 type-ui"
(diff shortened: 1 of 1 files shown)


g-a-v-i-n:
This is getting a a bit crowded and looks like Docs got dropped

jalehman:
I dropped docs intentionally because it's not really a primary action you need such valuable real estate for, and it's mentioned further down the page. I'd be curious to get @marisarow 's take though.

g-a-v-i-n:
docs is our most hit route after the index page

PR #1090, opened 2021-08-27
update hosting providers

jalehman:
We've got some new hosting providers on the block, and Get an Urbit ID appears to be offline.


g-a-v-i-n (approved):

marisarow:
What is the logic for the ordering? I'm assuming we'll need an unbiased logic for listing, such as alphabetical.

PR #1094, opened 2021-08-31
add assembly page, support multiday events

g-a-v-i-n:
- adds events page for Assembly
- supports rendering multiday events


urcades:
The added interfaces are looking really good, but I noticed some (collateral damage?) bugs in the rendering of the events pages, which appears to effect the currently-deployed site. Notably, "blog" page navigation renders perfectly.

Screen Shot 2021-08-31 at 12 05 01 PM

This is a screenshot from a 'mobile view' in my desktop browser, but this preview shows the same bugginess when viewed on an iphone

marisarow:
Curious why we aren't pushing full width or cropping the banner in closer for desktop view?
Screen Shot 2021-08-31 at 11 24 42 AM

urcades:
@marisarow that's a good catch, I didn't have a desktop big enough to catch this one — the banner appeared edge-to-edge on my laptop's screen + mobile viewport

g-a-v-i-n:
Nice catches, will fix both

g-a-v-i-n:
@urcades and @marisarow  

Resolved the following:
- Banner width issue
- Event tile rendering layout bug
- Event rendering on mobile

urcades (approved):
everything seems to be rendering nicely as far as I've been testing/checking

PR #1106, opened 2021-09-08
Fix issue in using/os/getting-started

timgiroux:
https://urbit.org/using/os/getting-started#moons

Problem:
   Markdown ```<key>``` becomes  HTML <code><key></key></code>

Solution:
   Remove the <> chars.


jalehman:
Thank you!

PR #1111, opened 2021-09-10
add zuse reference with json functions

tinnus-napbus:
- add new zuse section in /hoon/reference
- add a subset of zuse's functions relating to json

This is only a small subset of zuse relating to json decoding/encoding &
printing/parsing (plus a couple of others), but it's structured so more
zuse functions can be added as needed.


drbeefsupreme (changes requested):
Just add a table of contents page like https://urbit.org/docs/hoon/reference/stdlib/table-of-contents and make sure the arms show up in the search bar by adding entries for them at https://github.com/urbit/urbit.org/blob/master/lib/glossary.js

tinnus-napbus:
cool I've added the table of contents and glossary entries

drbeefsupreme (approved):

PR #1126, opened 2021-09-14
order future events by earliest event

jalehman:
With future events, I want to see the soonest event first rather than the latest one.


g-a-v-i-n:
Nice catch. What was `<GuestList />` supposed to do? Maybe I can help.

The other thing I'm noticing now is the guestlist on the small preview tiles can be too long - I think we might want to just display the host on the small preview cards.

Screen Shot 2021-09-14 at 11 24 21 AM

jalehman:
> The other thing I'm noticing now is the guestlist on the small preview tiles can be too long - I think we might want to just display the host on the small preview cards.

This is what the `GuestList` component was for, but couldn't quite get it working.

PR #1131, opened 2021-09-16
add json guide

### LINKED ISSUE #1093: json formatting and parsings docs
`enjs:format`, `dejs:format`, `en-json:html` and `de-json:html` aren't documented but I assume they'll be used very heavily by people who want to write apps. 

There ought to be reference docs a la the hoon standard lib ref as well as a short guide on basic usage.

they are in zuse rather than hoon.hoon so I'm not sure theoretically where they should go

Should there be a zuse section maybe? There's some other non-js formatting functions in zuse that would also be useful to document, idk. Or maybe just a json section in /docs/hoon/reference? Or arvo/concepts?

what do you think @drbeefsupreme

tinnus-napbus:
- explains $json type
- explains relevant libraries
- explains how you'd approach the issue in the context of a gall agent
- provides example of encoding/decoding a structure to $json
- provides example of a json conversions with a mark file

note this doesn't backtick every single instance of everything for readability purposes. I have backticked things when it seemed appropriate, typically when the thing was first introduced in the relevant section

note this links out to the zuse json documentation in #1111 and the mark file documentation in #769 so depends on those (if you don't want dead links anyway)

resolves #1093

### DIFF
--- a/content/docs/hoon/guides/json-guide.md
+++ b/content/docs/hoon/guides/json-guide.md
@@ -1,0 +1,354 @@
++++
+title = "JSON"
+weight = 4
+template = "doc.html"
++++
+
+If one is working on a Gall agent with any kind of web interface, it's likely one will encounter the problem of converting Hoon data structures to JSON and vice versa. This is what we'll look at in this document.
+
+Urbit represents JSON data with the `$json` structure defined in `lull.hoon`. You can refer to the [json type](#the-json-type) section below for details. Since JSON data on the web is typically encoded in text, Urbit has two functions in `zuse.hoon` - [+en-json:html](/docs/hoon/reference/zuse/2e_2-3#en-jsonhtml) for printing `$json` to a text-encoded form, and [+de-json:html](/docs/hoon/reference/zuse/2e_2-3#de-jsonhtml) for parsing text-encoded JSON to a `$json` structure.
+
+Since one typically wants `$json` data converted to some other `noun` structure or vice versa, Urbit has three collections of functions for this purpose, also in `zuse.hoon`. The first is [+enjs:format](/docs/hoon/reference/zuse/2d_1-5#enjsformat), which contains functions for converting various atoms and structures to `$json`. The second is [+dejs:format](/docs/hoon/reference/zuse/2d_6#dejsformat), which contains many JSON "reparsers" for converting `$json` data to atoms and other structures. The third is [+dejs-soft:format](/docs/hoon/reference/zuse/2d_7#dejs-softformat), which is largely the same as `+dejs:format` except its reparsers produce `unit`s which are null upon failure rather than simply crashing.
+
+The relationship between these types and functions look like this:
+
+![json diagram](https://m.tinnus-napbus.xyz/pub/json-diagram.svg)
+
+Note this diagram is a simplification - the `+dejs:format` and `+enjs:format` collections in particular are tools to be used in writing conversion functions rather than simply being used by themselves, but it demonstrates the basic relationships. Additionally, it would be less common to perform the printing/parsing directly - this would typically be handled implicitly by Eyre, though it may be necessary if one is requesting external JSON data via the web client vane Iris.
+
+### In practice
+
+A typical Gall agent will have a number of structures defined in a file in the `/sur` directory. These will define the type of data it expects to be `%poke`ed with, the type of data it will `%give` to subscribers, and the type of data its scry endpoints produce.
+
+If the agent is expected to only interact with local agents or agents on remote ships over Ames, it may just take and produce its structures with a `%noun` `mark`. If, however, it needs to talk to a web interface of some kind, it usually must take and produce `$json` data with a `%json` mark.
+
+It may be the case that its dealings with a web interface are totally distinct from its dealings with other agents. If so, the agent could just have separate scry endpoints, poke handling code, etc, that just directly take and produce `$json` data with a `%json` mark. In such a case one can include `$json` encoding/decoding functions directly in the agent or associated libraries, using the general techniques demonstrated in the [$json encoding and decoding example](#json-encoding-and-decoding-example) section below.
+
+If, on the other hand, one wants to be able to use the same basic structures in requests and responses, and the same code to deal with them, whether from the web or within Urbit, a different approach is necessary. Rather than taking or producing either `%noun` or `%json` marked data, custom `mark` files can be created for the relevant structures, which specify conversion methods for both `%noun` and `%json` marked data.
+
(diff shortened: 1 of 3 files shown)


content/docs/hoon/guides/json-guide.md
```diff
@@ -0,0 +1,363 @@
...
+  --
+++  grad  %noun
+--
+```
+
+The [Marks section](/docs/arvo/clay/marks/marks) of the Clay documentation covers `mark` files comprehensively and is worth reading through if one wants to write a mark file.
+
+In brief, a mark file contains a `door` with three arms. The door's sample type is the type of the data in question - in our case the `$user` structure. The `+grab` arm contains methods for converting _to_ our mark, and the `+grow` arm contains methods for converting _from_ our mark. The `+noun` arms are mandatory, and then we've added `+json` arms which respectively call the `+from-js` and `+to-js` functions from our `json-test.hoon` library. The final `+grad` arm defines various revision control functions, in our case we've delegated these to the `%noun` mark.
+
+From this mark file, Clay can build mark conversion `gate`s between the `%json` mark and our `%user` mark, allowing the conversion of `$json` data to a `$user` structure and vice versa.
```
drbeefsupreme:
```suggestion
From this mark file, Clay can build mark conversion gates between the `%json` mark and our `%user` mark, allowing the conversion of `$json` data to a `$user` structure and vice versa.
```

content/docs/hoon/guides/json-guide.md
```diff
@@ -0,0 +1,363 @@
...
+    p
+  { [p='email' q=[%s p='john.smith@example.com']]
+    [p='name' q=[%a p=~[[%s p='John'] [%s p='William'] [%s p='Smith']]]]
+    [p='username' q=[%s p='john456']]
+    [p='joined' q=[%n p=~.1631440078]]
+  }
+]
+```
+
+We'll also build our library so we can use its types from the Dojo:
```
drbeefsupreme:
```suggestion
We'll also build our library so we can use its types from the dojo:
```

content/docs/hoon/guides/json-guide.md
```diff
@@ -0,0 +1,363 @@
...
+> =user-lib -build-file %/lib/json-test/hoon
+```
+
+Now we can ask Clay to build a mark conversion gate from a `%json` mark to our `%user` mark. We'll use a scry with a `%f` `care` which produces a static mark conversion gate:
+
+```
+> =json-to-user .^($-(json user:user-lib) %cf /===/json/user)
+```
+
+Let's try convert our `$json` to a `$user` structure with our new mark conversion gate:
```
drbeefsupreme:
```suggestion
Let's try converting our `$json` to a `$user` structure with our new mark conversion gate:
```

drbeefsupreme (changes requested):
LGTM. I think it would also be good to mention this guide at the top of `/hoon/guides/parsing.md`, like "if you're looking for a guide to parse JSON specifically, see here"

basilesportif (changes requested):
OK, couple things left to do:

- note that your `from-js` gate is a parser, but also that the gates `ot` uses are JSON parsers as well. It's important to convey that recursive nature.
- decoding JSON is the thing that gives people the most trouble. So there should be more about patterns with `ot`/`of`. Link to or copy relevant source from the codebase for examples?
- should have explanation/examples for decoding optional fields
- cover `pairs` and the simple encoding primitives more

tinnus-napbus:
ok @timlucmiptev I added a new section that covers `ou` and `of`, and I changed the enjs example to use pairs and added a note about frond. I also added a note emphasising the recursive nesting of dejs functions

basilesportif:
That works for me.

content/docs/hoon/guides/json-guide.md
```diff
@@ -0,0 +1,453 @@
...
+#### Converting from `$json`
+
+Here's our arm that converts `$json` to our `$user` structure:
+
+```hoon
+++  from-js
+  =,  dejs:format
+  |=  jon=json
+  ^-  user
+  %.  jon
```
liam-fitzgerald: This is not strictly necessary, cause these runes are effectively a no-op. I'd scrap these three lines and replace it with  `^-  $-(json user)` cast, to make it clear that the `dejs` reparsers are higher order functions

content/docs/hoon/guides/json-guide.md
```diff
@@ -0,0 +1,453 @@
...
+Finally, let's see how that looks as JSON encoded in text:
+
+```
+> (crip (en-json:html (user-to-json usr)))
+'{"joined":1631440078,"username":"john456","name":["John","William","Smith"],"email":"john.smith@example.com"}'
+```
+
+Usually (though not in all cases) these mark conversions will be performed implicitly by Gall or Eyre and you'd not deal with the mark conversion gates directly, but it's still informative to see them work explicitly.
+
+## Further reading
```
liam-fitzgerald, line 499: Worth linking directly to `scot` imo, given its prevalence in JSON conversion.

content/docs/hoon/guides/json-guide.md
```diff
@@ -0,0 +1,453 @@
...
+
+For `joined`, we've used the [`+du`](/docs/hoon/reference/zuse/2d_6#dudejsformat) function, which converts a Unix seconds timestamp in a `$json` number to a `@da` (it's basically the inverse of the `+sect:enjs:format` we used earlier).
+
+Notice how `+ot` takes in other `+dejs` functions in its argument. One of its arguments includes the `+at` function which itself takes in other `+dejs` functions. There are several `+dejs` functions like this that allow complex nested JSON structures to be decoded. For other examples of common `+dejs` functions like this, see the [More `+dejs`](#more-dejs) section below.
+
+There are dozens of different functions in [`+dejs:format`](/docs/hoon/reference/zuse/2d_6) that will cover a great many use cases. If there isn't a `+dejs` function for a particular case, you can also just write a custom function - it just has to take `$json`. Note there's also the [`+dejs-soft:format`](/docs/hoon/reference/zuse/2d_7) functions - these are similar to `+dejs` functions except they produce `unit`s rather than simply crashing if decoding fails.
+
+## More `+dejs`
```
liam-fitzgerald, line 242: Nothing on `+su`? Imo one of the most important decoders, because there's a lot of pressure on JSON objects to be somewhat stringly-typed (keys must be strings, strings don't break pointer equality unexpectedly).

liam-fitzgerald (commented):
Looks good! Some minor nits, but looks terrific broadly. Thank you for this

tinnus-napbus:
@liam-fitzgerald cool I've addressed your comments and added a +su section

liam-fitzgerald (approved):
LGTM!

PR #1173, opened 2021-09-30
dist-faq: update %cz hashes

Fang-:
As per latest updates.


eamsden (approved):
I added a commit to change the timestamp, which we should do whenever we update the cz hashes at least. Otherwise looks fine

PR #1177, opened 2021-10-01
change binzod to litzod, %base install

drbeefsupreme:
some people seem to be having connectivity trouble with ~binzod, also fixed the %base install command


Fang-:
Worth mentioning that ~litzod has spawned _many_ more planets than the other four "primary" ~zod stars. While not all those planets are in active use, we should consider avoiding additional load on ~litzod, and point people to the other stars instead.

The `|install ~binzod %base` might even be the source of the connectivity reports here. Ships' base desk is not public, and so install from that will just not work.

PR #1178, opened 2021-10-02
banner change

g-a-v-i-n:
- replace release banner with new banner


marisarow:
We need punctuation after Assembly 2021 (missing a period)

Network update is live. Learn more at Assembly 2021.

PR #1184, opened 2021-10-04
add a guide to strings in hoon

tinnus-napbus:
- the main string types
- explain syntax
- string interpolation
- printing/parsing atom auras
- conversions


content/docs/hoon/guides/strings.md
```diff
@@ -0,0 +1,502 @@
...
+
+```hoon
+~.abc-123.def_456~ghi
+```
+
+### `term`
+
+A [`term`](/docs/hoon/reference/stdlib/2q#term) has an aura of
+`@tas`, and is a subset of a [`knot`](#knot). It only allows lower-case letters,
+numbers, and hyphens. Additionally, the first character cannot be a hyphen. This
```
Fang-:
```suggestion
numbers, and hyphens. Additionally, the first character cannot be a hyphen or number. This
```

tinnus-napbus: oh yeah true. I thought it was ok cos the dojo didn't balk at it but it fails sane

basilesportif (changes requested):
Lgtm with one thing: you should give a quick overview of when you'd want tape vs. the cord/knot/term family.

tinnus-napbus:
@timlucmiptev I've added a section comparing tapes vs cords etc and noting when you'd use them. I'm kinda assuming the different use cases here, let me know if I'm missing something

basilesportif (approved):
LGTM

Issue #1185, opened 2021-10-04
glossary.js entries eg stdlib functions appear then disappear in search results


tinnus-napbus:
If you search for eg `scot`, it should have the glossary.js entry as the first result and a link to the relevant stdlib docs page and section. For some reason when you type eg `scot` it appears for half a second then disappears, and you have to backspace the last character and re-enter it to get the result back.

tinnus-napbus:
fixed by #1189

PR #1166, opened 2021-09-29
docs: fix code block formatting

jamesacklin:
Adds some styles to `code` blocks wrapped in `pre`. Also allows for overscroll on wide blocks with small main column widths.

Before:
image

After: 
image

Does not address the endemic Markdown parsing issues, but at least treats this visible symptom.

Fixes urbit/urbit.org#1139

Issue #1160, opened 2021-09-27
image embeds not appearing when they're html tags in markdown files


tinnus-napbus:
See:

https://urbit.org/docs/userspace/graph-store/validator-walkthrough

vs:

https://github.com/urbit/urbit.org/blame/master/content/docs/userspace/graph-store/validator-walkthrough.md#L26-L28

I assume this is related to the other issues with codeblocks and section links and such

g-a-v-i-n:
yes, I think this is related to the markdown chain

tinnus-napbus:
fixed by #1189

Issue #1153, opened 2021-09-22
Superscript tags not rendering as intended


jtobin:
28, 216, and 232 in the following should render as 2^8, 2^16, and 2^32:

![](https://jtobin.s3-ap-southeast-2.amazonaws.com/Screen-Shot-2021-09-22-15-19-49.85.png)

The HTML tags seem correct, so I guess this might be a CSS issue?

```
Urbit IDs are distributed by a sponsorship tree. At the top of the tree
are 2<sup>8</sup> (256) galaxies. Each galaxy issues 2<sup>8</sup>
stars, making a total of 2<sup>16</sup> (65K). Stars then each can issue
2<sup>16</sup> planets, making for 2<sup>32</sup> (~4B). As you might
expect, each planet issues 2<sup>32</sup> moons.
```

ohAitch:
Nonsense, there are only 232 total planets XD

g-a-v-i-n:
nice catch

tinnus-napbus:
Fixed by #1189

PR #1189, opened 2021-10-06
Markdown fixes

arthyn:
I think this fixes most of the problems with the markdown chain. Stated in this email:

<blockquote>
1. Section links are all prepended with user-content so urbit.org/foo#bar becomes urbit.org/foo#user-content-bar. This breaks all the links to sections throughout the docs. Issue: https://github.com/urbit/urbit.org/issues/1141

2) Tall-form codeblocks were rendering as though they were inline codeblocks. See issue https://github.com/urbit/urbit.org/issues/1139 - this was fixed in PR https://github.com/urbit/urbit.org/pull/1166 but as I understand it, it's a stop-gap fix that hasn't addressed the underlying cause.

3) Syntax highlighting's not working

4) search results for glossary.js entries acting weird, see https://github.com/urbit/urbit.org/issues/1185

5) there are various cases of html elements embedded in markdown files not rendering, eg images in html tags don't show. See https://github.com/urbit/urbit.org/issues/1160 for instance. The linked examples in the issue have been fixed by changing to markdown image embeds but it's still an issue and I think other html tag cases like eg https://github.com/urbit/urbit.org/issues/1153 have the same problem

6) html inside codeblocks is having bits of the tags stripped, see the html codeblock in
https://urbit-org-kta5u1772-urbit.vercel.app/docs/userspace/dist/guide#user-content-create-files-for-glob for example 
</blockquote>

Oof remark is hard to find information on. We have one remaining problem which is that fenced code blocks without a language spit out double escaped html entities like this(&#38;gt;) so we have to run an HTML decoder on all our markdown output which breaks blocks of HTML inside code fences like on tinnus-napbus' distribution guide.

The only solution I can think of is somehow give all our code blocks some default language like bash, find a quick way to do find and replace that won't screw everything up, or do it manually lol.


g-a-v-i-n:
Amazing work.

My personal opinion is that we should steer away from HTML in markdown and handle the UX of Getting Started etc in a specific page designed for onboarding. Having HTML embedded in markdown is a convenient escape hatch but I strongly prefer not to rely on it going forward.

arthyn:
That sounds like a good plan, then we could rip out the `allowdangerously` and `rehype-raw` stuff. We could also implement MDX if we want more advanced components like this in MD later on.

g-a-v-i-n:
Cool. I would still like to keep the content we have that is currently using html but would like some kind of transition plan for it.

jalehman:
@g-a-v-i-n is there anything preventing merging this? or is it good to go?

g-a-v-i-n:
Looks fine to me, but might break HTML-in-markdown per Hunter's comment
above
—
~ridlur-figbud
https://urbit.org

arthyn:
Oh sorry, HTML in markdown is fine, what I was referencing was HTML in code examples a la triple backticks. This PR actually fixes HTML in markdown so we get the nice tabs in stuff like the CLI install.

g-a-v-i-n:
Awesome

PR #1207, opened 2021-10-12
update operators manual wrt dist

tinnus-napbus:
- change all reference to home desk to base
- update relevant explanations of desks
- change ota instructions
- update outdated instructions
- remove deprecated instructions
- various other fixes

this edits /using/os only, specifically basics, shell and filesystem. Doesn't touch id section.

This needs a much more substantial overhaul but for now its information should atleast be technically correct


drbeefsupreme (approved):
LGTM

PR #1208, opened 2021-10-13
Ed/add-audits

urcades:
Adding a placeholder page + footer item for a forthcoming "Audits" page for Urbit.org

### DIFF
--- a/components/Footer.js
+++ b/components/Footer.js
@@ -62,6 +62,13 @@ export default function Footer() {
                 <a className="mt-2 type-bold text-wall-500">Urbit OS</a>
               </div>
             </Link>
+
+            <Link href="/faq">
+              <div>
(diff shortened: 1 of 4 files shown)


drbeefsupreme:
@urcades is this PR a draft?

If not, just listing the audits here https://urbit-org-6cxistc51-urbit.vercel.app/audits is fine but we can remove the blurb.

This would also be a good page to link to from the FAQ.

urcades:
@drbeefsupreme for a little bit longer, yes — that blurb is meant to embody a bit of flavor text to describe the audits page and its contents.

That being said, on second thought, flavor text is secondary to the list itself, so perhaps we should just push it. I'll go ahead and do this, and we can add the description in another mini PR, along with a link in the FAQ.

PR #1217, opened 2021-10-16
Correct for html rendering of <p> in ;+ by escaping

rabsef-bicrym:
previously, <p> was entered in markdown as '<p>' which was, on urbit.org, rendering into a blank paragraph in code highlighting (sorta amusing, really).

By using <code>&lt;p&gt;</code>, the characters escape parsing by the browser and are rendered as characters, not a paragraph tag.


ohAitch:
Did the double-markdowning not get fixed? Like, how many examples in the [markdown spec](https://spec.commonmark.org/0.30/) currently get misrendered, as a proxy for "existing content we haven't noticed breakage in yet"?

ohAitch:
(This one is [example 343](https://spec.commonmark.org/0.30/#example-343) more or less)

PR #1229, opened 2021-10-27
make environment setup guide more comprehensive

tinnus-napbus:
- more detail about creating and managing fake ships
- explain managing desks in light of software dist changes
- best practices for structuring projects
- restructure

### DIFF
--- a/content/docs/development/environment.md
+++ b/content/docs/development/environment.md
@@ -267,7 +267,7 @@ and copy the actual files.
 Now you can just add a `sys.kelvin` file:
 
 ```
-echo "[%zuse 420]" > mydesk/sys.kelvin
+echo "[%zuse 419]" > mydesk/sys.kelvin
 ```
 
 And you'll be able to mount the desk with `|commit %mydesk`.
--- a/content/docs/development/environment.md
+++ b/content/docs/development/environment.md
@@ -8,25 +8,7 @@ template = "doc.html"
 This guide covers best practices for preparing your environment to develop
 within the Urbit ecosystem.
 
-## Creating a development ship
-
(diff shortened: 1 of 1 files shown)


content/docs/development/environment.md
```diff
@@ -61,4 +43,271 @@ with [hoon-mode.el](https://github.com/urbit/hoon-mode.el).
...
+│   └── txt.hoon
+└── sys.kelvin
+```
+
+`sys.kelvin` specifies the kernel kelvin version with which the desk is
+compatible. You can copy it across from the `%base` desk, or just run the
+following in the terminal from within the desk directory:
+
+```sh
+echo "[%zuse 420]" > sys.kelvin
```
drbeefsupreme: OTA updating to 419 is going out today
```suggestion
echo "[%zuse 419]" > sys.kelvin
```

content/docs/development/environment.md
```diff
@@ -61,4 +43,271 @@ with [hoon-mode.el](https://github.com/urbit/hoon-mode.el).
...
+```
+
+Note you have to use `cp -rL` rather than just `cp -r` because the
+`symbolic-merge.sh` script creates symlinks, so the `L` flag is to resolve them
+and copy the actual files.
+
+Now you can just add a `sys.kelvin` file:
+
+```
+echo "[%zuse 420]" > mydesk/sys.kelvin
```
drbeefsupreme:
```suggestion
echo "[%zuse 419]" > mydesk/sys.kelvin
```

drbeefsupreme (changes requested):

drbeefsupreme (approved):

PR #1230, opened 2021-10-29
Gall add data types and restructure

tinnus-napbus:
- Gall vane documentation is moved out of userspace and back into
  arvo/gall
- add data types document to gall vane docs
- move gall guide to top level of userspace docs
- fix all links

It's cleaner if actual Gall vane documentation follows the same
structure as other vanes, rather than being nested under userspace.
Ultimately the api doc should be redone to specifically cover gall vane
tasks like the other vanes, at the moment it duplicates a lot of info
with the data types section and the overview section. Will do at some
stage but not done for now.

The data types section is so there's a canonical gall types reference
that the new gall guide can link out to.


drbeefsupreme:
LGTM

PR #1232, opened 2021-11-01
Fix ambiguity in hoon-syntax.md

danpovey:
Change "have another irregular form" -> "contain another irregular form".  "have another irregular form" can too easily be read as "admit another irregular form", which is incorrect.


drbeefsupreme:
Thanks!

PR #1085, opened 2021-08-26
=* deferred expression

drbeefsupreme:
Its not a macro. Interestingly, the blurb in the search bar already knew this. I have a vague recollection of fixing this over a year ago but can't find the PR, maybe it somehow got reverted at some point. EDIT: nope see below

PR #1254, opened 2021-11-23
lustar alias -> deferred expression

drbeefsupreme:
In #1085 we decided what `=*` actually does to be "deferred expression"
rather than alias or macro. Lustar never got the same treatment, so this fixes
that up.

### DIFF
--- a/content/docs/hoon/reference/rune/lus.md
+++ b/content/docs/hoon/reference/rune/lus.md
@@ -135,7 +135,7 @@ Any Hoon expression, `q`, may be used to define the arm computation.
 
 ### `+*` "lustar"
 
-Defines aliases within doors.
+Defines deferred expressions within doors.
 
 ##### Syntax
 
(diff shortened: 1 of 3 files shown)


ohAitch:
Lazy evaluation :P

drbeefsupreme:
That would be another good thing to throw in there. Since @belisarius222 suggested the term deferred evaluation, is there any significant difference between this notion and lazy evaluation? Incredibly pedantic hair splitting encouraged.

Fang-:
Oh man, here we go again.

Generally "lazy evaluation" is understood to "evaluate when first needed", instead of "reevaluate whenever needed", right?

ohAitch:
That's a fair point, though I'm not sure if "deferred" evaluation is clearer on that front.

drbeefsupreme:
Just gonna merge this to keep things consistent with the `=*` entry and the new Gall guide. I feel like this is just gonna be a contentious topic until "tistar" is added to the dictionary and we can just start calling it that.

PR #1265, opened 2021-12-06
Typo on example planet

alentodorov:
I think the `sampel-planet` should've been named `sample-planet`.


alentodorov:
I aplogize for this PR. I should've read more about Urbit to understand this is part of the lingo.

PR #1273, opened 2021-12-09
json-guide: add note about foo+bar syntax

### LINKED ISSUE #1264: JSON guide should explain lus sugar
The JSON guide jumps from examples that use things like `[%username so]` to the `username+so` format without any explanation. Should have a short explanation of that sugar.  Apologies if the explanation is there and I missed it.

tinnus-napbus:
resolves #1264


tinnus-napbus:
y no vercel deploy...

PR #1279, opened 2021-12-15
Fix CSS width bug on several pages

raglep-ragret:
urbit.org has a little CSS bug that's making the page width wider than it needs to be. The width extends beyond the screen, forcing a bottom scrollbar even though there is no content in the blank area:

image

This appears to affect all non-Sidebar-layout pages. Luckily, it's all caused by a simple bug in `Container`. This bug's a classic: [your scrollbar is part of your viewport](https://stackoverflow.com/questions/23367345/100vw-causing-horizontal-overflow-but-only-if-more-than-one/23367686), so if there is a vertical scrollbar in a `100vw` element, that element will end up one scrollbar wider than the screen's visible area. In in this case, it's easy to fix- we can just add `max-w-full`, which fixes the problem:

image

(While I was in there, I noticed some repeated `w-screen` classes, and figured I might as well fix those too.)

### DIFF
--- a/components/Container.js
+++ b/components/Container.js
@@ -1,7 +1,7 @@
 // Ensures the root container is always 100vw, min 100vh, and centers all children along the y-axis
 export default function Container({ children }) {
   return (
-    <div className="flex flex-col min-h-screen w-screen items-center">
+    <div className="flex flex-col min-h-screen w-screen max-w-full items-center">
       {children}
     </div>
   );
--- a/pages/community/[[...slug]].js
+++ b/pages/community/[[...slug]].js
@@ -66,7 +66,7 @@ export default function UsingLayout({ posts, data, params, search, markdown }) {
         <title>{data.title} • Community • urbit.org</title>
         {Meta(data)}
       </Head>
-      <div className="flex w-screen h-screen min-h-screen w-screen sidebar">
+      <div className="flex h-screen min-h-screen w-screen sidebar">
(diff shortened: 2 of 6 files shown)


pages/community/[[...slug]].js
```diff
@@ -66,7 +66,7 @@ export default function UsingLayout({ posts, data, params, search, markdown }) {
         <title>{data.title} • Community • urbit.org</title>
         {Meta(data)}
       </Head>
-      <div className="flex w-screen h-screen min-h-screen w-screen sidebar">
+      <div className="flex h-screen min-h-screen w-screen sidebar">
```
Comment (raglep-ragret) [inline pages/community/[[...slug]].js:69]: Changes from here on down simply remove one duplicate `w-screen` per file.

components/Container.js
```diff
@@ -1,7 +1,7 @@
 // Ensures the root container is always 100vw, min 100vh, and centers all children along the y-axis
 export default function Container({ children }) {
   return (
-    <div className="flex flex-col min-h-screen w-screen items-center">
+    <div className="flex flex-col min-h-screen w-screen max-w-full items-center">
```
raglep-ragret, line 4: Here's where the action is. Just adding this `max-w-full` should be all it takes to fix the width.

matildepark (approved):
Looks fine on preview, mobile layouts unaffected, this seems fine to merge

PR #1281, opened 2021-12-16
Update ship.md

e-shee:
changing peer to pier @nerveharp @crrncllns


ynx0:
might wanna check spelling

e-shee:
fixes one typo...makes another

tinnus-napbus:
"peer" is correct in this context - a pier is the folder containing the event log, checkpoint and such, and isn't "on the network"

Issue #1025, opened 2020-08-25
add +cook and +next to parsing tutorial


drbeefsupreme:
Per @Fang- 's suggestion

PR #1306, opened 2022-01-14
minor edits to parse tutorial, add +cook and +next

drbeefsupreme:
Addresses #1025 and adds some more links and examples. Also fixes an embarassing
nonsensical error (showing a non-working recursive parse, then claiming that a
totally different parser was the fixed version of it) that probably resulted
from my own sloppy editing when I first wrote this.


tinnus-napbus (approved):
LGTM

PR #1308, opened 2022-01-16
Add instructions for  Node.js 17 incompatibility

rabsef-bicrym:
[Node.js v17 comes with webpack 4 which has some incompatibilities in cryptography suites](https://stackoverflow.com/questions/69665222/node-js-17-0-1-gatsby-error-digital-envelope-routinesunsupported-err-os).  Installing and running urbit.org can still be accomplished using Node.js v17+ using the added commands, at least until the project is upgraded to webpack 5.

This is my limited understanding based on my limited experience w/ JavaScript and some searching of stackoverflow. If I am mistaken, my apologies for this commit - this did, however, work for me to build locally without changing node versions.

### DIFF
--- a/README.md
+++ b/README.md
@@ -16,6 +16,39 @@ yarn dev
 
 Finally, open [http://localhost:3000](http://localhost:3000) with your browser to see the result.
 
+### Errors with Digital Envelope Routines
+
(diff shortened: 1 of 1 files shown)


jalehman (approved):

matildepark:
Er, but this isn't a Gatsby site though? Those instructions here can cause confusion... I think best case here is to pin us to node 16 with a `.nvmrc`.

PR #1322, opened 2022-02-03
Fix broken links

dnmfarrell:


vercel.json
```diff
@@ -25,7 +25,7 @@
     },
     {
       "source": "/breach",
-      "destination": "/reset",
+      "destination": "/using/id/guide-to-resets",
```
tinnus-napbus:
```suggestion
      "destination": "/reset",
```

tinnus-napbus: this was used for the page about the network breach at the end of 2020, not for info about breaching, so should probably be left as-is (although that page doesn't work for some reason, dunno why though. I think it's missing a corresponding .js file)

tinnus-napbus (changes requested):

vercel.json
dnmfarrell: Gotcha, I've undone that change.

tinnus-napbus (approved):
LGTM

PR #1331, opened 2022-02-15
Add bounties page and submission form

jalehman:


mikeosborne (approved):
Looks good to me. I like how you used a format consistent with the Apprenticeship page. My only suggestion has to do with the Airtable form - I'll make changes there.

PR #1342, opened 2022-02-21
add sail guide

tinnus-napbus:
This is an update, edited, revised version of the old sail guide from 2019 (or whenever it was). The framing of writing whole documents in sail is replaced by a framing of using it in agents to render front-ends.

### DIFF
--- a/content/docs/glossary/sailudon.md
+++ b/content/docs/glossary/sailudon.md
@@ -8,6 +8,7 @@ category = "arvo"
 
 **Sail** and **Udon** are domain specific languages for Hoon.
 
-Sail is used to express XML data structures, which are used for rendering webpages, such as when browsing [collections](/docs/glossary/collection) in [Landscape](/docs/glossary/landscape).
+Sail is used to express XML data structures, which are used for rendering
+webpages. See the [Sail Guide](/docs/hoon/guides/sail) for more details.
 
 Udon is similar to Markdown, a simple language for writing documents. Udon is stored in [Clay](/docs/glossary/clay) and rendered using [Ford](/docs/glossary/ford).
--- a/content/docs/glossary/sailudon.md
+++ b/content/docs/glossary/sailudon.md
@@ -8,7 +8,7 @@ category = "arvo"
 
 **Sail** and **Udon** are domain specific languages for Hoon.
 
-Sail is used to express XML data structures, which are used for rendering
+Sail is used to express XML data structures, which are commonly used for rendering HTML
 webpages. See the [Sail Guide](/docs/hoon/guides/sail) for more details.
 
 Udon is similar to Markdown, a simple language for writing documents.
--- a/content/docs/glossary/sailudon.md
+++ b/content/docs/glossary/sailudon.md
@@ -11,4 +11,4 @@ category = "arvo"
 Sail is used to express XML data structures, which are used for rendering
 webpages. See the [Sail Guide](/docs/hoon/guides/sail) for more details.
 
-Udon is similar to Markdown, a simple language for writing documents. Udon is stored in [Clay](/docs/glossary/clay) and rendered using [Ford](/docs/glossary/ford).
+Udon is similar to Markdown, a simple language for writing documents.
--- a/content/docs/hoon/guides/sail.md
+++ b/content/docs/hoon/guides/sail.md
@@ -59,8 +59,8 @@ It’s easy to see how Sail can directly translate to HTML:
 
 In Sail, tag heads are written with the tag name prepended by `;`. Unlike in
 HTML, there are different ways of closing tags, depending on the needs of the
-tag. One of the nice things about Hoon is that you don’t have to be constantly
-closing expressions; Sail inherits this convenience.
+tag. One of the nice things about Hoon is that you don’t have to constantly
+close expressions; Sail inherits this convenience.
 
 ### Empty
 
--- a/content/docs/hoon/guides/sail.md
+++ b/content/docs/hoon/guides/sail.md
@@ -315,16 +315,47 @@ elements, converting characters such as `>` to HTML entities. For example:
 
 ### Marks
 
-The mark for a raw `$manx` is a `%hymn`. This mark is included in the `%base`
-desk. The mark for a rendered HTML cord is `%html`. The `%hymn` mark includes
-conversion methods to both `%html` and `%mime`. The `%html` mark includes
-conversion methods to `%mime` and `%hymn`.
+There are a few different HTML and XML related marks, so it can be a bit
+confusing. We'll look at the ones you're most likely to use.
 
-If you want to produce an HTML document through Eyre's scry interface, for
-example, you can either produce a `$manx` with a `%hymn` mark and let Eyre
-convert it to `%html`, or else you can call `++en-xml:html` and then `++crip` on
-the product of your Sail expressions and produce a cord with an `%html` mark
-directly.
+#### `%html`
+
+- Type: `@t`
+
+This mark is used for HTML that has been printed as text in a cord. You may wish
+to return this mark when serving pages to the web. To do so, you must run the
+`$manx` produced by your Sail expressions through `++en-xml:html`, and then run
+the resulting `tape` through `++crip`.
+
+#### `%hymn`
+
+- Type: `$manx`
+
+The `%hymn` mark is intended to be used for complete HTML documents - having an
+`&lt;html>` root element, `&lt;head>`, `&lt;body>`, etc. This isn't enforced on
+the type level but it is assumed in certain mark conversion pathways. Rather
+than calling `++en-xml:html` on the `$manx` produced by your Sail expressions
+and returning an `%html` mark, you may wish to return the `$manx` directly as a
+`%hymn` and let Eyre handle converting it to `%html` before giving it to web
+clients.
+
+#### `%elem`
+
+- Type: `$manx`
+
+The type of the `%elem` mark is a `$manx`, just like a `%hymn`. While `%hymn`s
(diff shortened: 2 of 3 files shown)


content/docs/glossary/sailudon.md
```diff
@@ -8,6 +8,7 @@ category = "arvo"
 
 **Sail** and **Udon** are domain specific languages for Hoon.
 
-Sail is used to express XML data structures, which are used for rendering webpages, such as when browsing [collections](/docs/glossary/collection) in [Landscape](/docs/glossary/landscape).
+Sail is used to express XML data structures, which are used for rendering
+webpages. See the [Sail Guide](/docs/hoon/guides/sail) for more details.
 
 Udon is similar to Markdown, a simple language for writing documents. Udon is stored in [Clay](/docs/glossary/clay) and rendered using [Ford](/docs/glossary/ford).
```
sigilante: "Udon documents are stored in Clay and built using the `++ford` arm."  If this sentence is even necessary, which I rather think it is not.

content/docs/hoon/guides/sail.md
```diff
@@ -0,0 +1,495 @@
++++
+title = "Sail (HTML)"
+weight = 6
+template = "doc.html"
++++
+
+Sail is a domain-specific language for composing HTML structures in Hoon.
```
sigilante: "HTML (and XML) structures"

content/docs/hoon/guides/sail.md
```diff
@@ -0,0 +1,495 @@
...
+Front-ends for Urbit apps are often created and uploaded separately to the rest
+of the code in the desk. Sail provides an alternative approach, where front-ends
+can be composed directly inside agents.
+
+This document will walk through the basics of Sail and its syntax.
+
+## Basic example
+
+It’s easy to see how Sail can directly translate to HTML:
```
sigilante, line 19: We could do this as a side-by-side table.

content/docs/hoon/guides/sail.md
```diff
@@ -0,0 +1,495 @@
++++
+title = "Sail (HTML)"
+weight = 6
+template = "doc.html"
++++
+
+Sail is a domain-specific language for composing HTML structures in Hoon.
+Front-ends for Urbit apps are often created and uploaded separately to the rest
+of the code in the desk. Sail provides an alternative approach, where front-ends
+can be composed directly inside agents.
```
sigilante: "Like everything else in Hoon, a Sail document is a binary tree data structure, just one produced by a specialized markup language within Hoon."

content/docs/hoon/guides/sail.md
```diff
@@ -0,0 +1,495 @@
...
+    </p>
+  </body>
+</html>
+```
+
+## Tags and Closing
+
+In Sail, tag heads are written with the tag name prepended by `;`. Unlike in
+HTML, there are different ways of closing tags, depending on the needs of the
+tag. One of the nice things about Hoon is that you don’t have to be constantly
```
sigilante: "you don't have to constantly close"

content/docs/hoon/guides/sail.md
```diff
@@ -0,0 +1,495 @@
...
+
+If we want to write a string with no tag at all, then we can prepend
+those untagged lines with `;` and then a space:
+
+```hoon
+;body
+  ;h1: Welcome!
+  ; Hello, world!
+  ; We’re on the web.
+==
```
sigilante, line 111: Part of me wants to suggest the convention of

```
==  :: body
```

content/docs/hoon/guides/sail.md
```diff
@@ -0,0 +1,495 @@
...
+
+Produces:
+
+```
+<main>
+  <p>This is the first node.</p>
+  <p>This is the second.</p>
+  <p>Here is the last one.</p>
+</main>
+```
```
sigilante, line 455: Let's have a section of "Good Examples" with links to projects like %pals and %gora.

content/docs/hoon/guides/sail.md
```diff
@@ -0,0 +1,495 @@
...
+> =z ;p(class "\"><script src=\"example.com/xxx.js"): <h1>FOO</h1>
+
+> (crip (en-xml:html z))
+'<p class="&quot;&gt;&lt;script src=&quot;example.com/xxx.js">&lt;h1&gt;FOO&lt;/h1&gt;</p>'
+```
+
+### Marks
+
+The mark for a raw `$manx` is a `%hymn`. This mark is included in the `%base`
```
sigilante: I think a little summary table of manx/marl/hymn etc. could be helpful.

tacryt-socryp: Wouldn't it make sense to just go ahead and change the mark `%hymn` to a name that makes sense like `%manx` or `%xml` lol?

content/docs/glossary/sailudon.md
sigilante: @tinnus-napbus do you want to put `;>` in here as well?

content/docs/hoon/guides/sail.md
tinnus-napbus: I'll redo this sections cos it's a bit inaccurate. Technically a `%hymn` is an HTML document, and an `%elem` is any xml node. They're both manxes in terms of their type but `%hymn` is intended to have a proper html > head & body structure while an `%elem` is not. There is already an `%xml` mark, along with `%htm`, `%html` `%x-htm`, `%x-htm-elem`. The `%hymn` mark is used throughout these in terms of mark conversions as well as in other source code so can't really change without disentangling a whole bunch of mark conversion pathways and changing other stuff

content/docs/glossary/sailudon.md
```diff
@@ -8,6 +8,7 @@ category = "arvo"
 
 **Sail** and **Udon** are domain specific languages for Hoon.
 
-Sail is used to express XML data structures, which are used for rendering webpages, such as when browsing [collections](/docs/glossary/collection) in [Landscape](/docs/glossary/landscape).
+Sail is used to express XML data structures, which are used for rendering
```
Fang-:
```suggestion
Sail is used to express XML data structures, which are commonly used for rendering HTML
```

content/docs/hoon/guides/sail.md
Fang-: This doesn't look resolved to me. Sail is _actually_ XML, it just so happens that HTML fits within that. The `sailudon.md` gets it right.

content/docs/hoon/guides/sail.md
```diff
@@ -0,0 +1,495 @@
...
+
+```
+<h1>The title</h1>
+```
+
+### Nested
+
+To nest tags, simply create a new line. Nested tags need to be closed with `==`,
+because they expect a list of sub-tags. If we nest lines of plain text with no
+tag, the text will be wrapped in a `&lt;p>` tag.
```
Fang-: Notably this also does some other things, like turning `%hello` into `<code>%hello</code>`. It's weird and we may not want to keep doing this, but it's where sail's at rn.

content/docs/hoon/guides/sail.md
```diff
@@ -0,0 +1,495 @@
...
+```
+<h1 class="text-blue">Title</h1>
+```
+
+If you want a class name that contains a space, you will need to use use the
+syntax of a generic attribute:
+
+```hoon
+;div(class "logo inverse");
+```
```
Fang-: Not really. This is just the equivalent of `div.logo.inverse`. These attribute shorthands stack & chain, assuming you put them in the right order.

content/docs/hoon/guides/sail.md
```diff
@@ -0,0 +1,495 @@
...
+  '''
+```
+
+And then your style tag might look like:
+
+```hoon
+;style: {style}
+```
+
+A cord is used rather than a tape so you don't need to escape double-quotes. The
```
Fang-: It'd be a `"""`, so single double quotes are fine. You really care about the `{` though.

content/docs/hoon/guides/sail.md
```diff
@@ -0,0 +1,495 @@
...
+
+> (en-xml:html x)
+"<p>foobar</p>"
+
+> (crip (en-xml:html x))
+'<p>foobar</p>'
+```
+
+You may wish to do this in your agent directly, but you can also leave mark
+conversion to Eyre.
```
Fang-: What flow is this referring to?

content/docs/hoon/guides/sail.md
```diff
@@ -0,0 +1,495 @@
...
+```
+
+In order to nest it inside another Sail element, it must be preceeded with a
+`;+` rune or similar, it cannot be used directly. For example:
+
+```hoon
+;p
+  ;+  ;/  "foobar"
+==
+```
```
Fang-: Some `?:` might make for a less trivial/more illustrative example.

content/docs/hoon/guides/sail.md
```diff
@@ -0,0 +1,495 @@
...
+
+```
+<p><b>2 </b>is an <b>even </b>number.</p>
+```
+
+```
+<p><b>12345 </b>is an <b>odd </b>number.</p>
+```
+
+### `;/` Micfas
```
Fang-: I would move this one to be the last rune mentioned, just because the other ones are so similar to each other, this one kinda stands out.

Fang- (commented):
Looks good! Couple comments, nothing too egregious.

content/docs/hoon/guides/sail.md
Fang-: Oops okay, after refresh it's actually resolved, but I could still nitpick about getting the "xml means html" ordering right.

PR #1345, opened 2022-02-23
add roller tutorial

drbeefsupreme:
This adds a short tutorial on how to set up a L2 roller. There is also a small
change to the glossary page on rollups to point to the layer 2 guide for planets.

### DIFF
--- a/content/docs/azimuth/l2/roller-tutorial.md
+++ b/content/docs/azimuth/l2/roller-tutorial.md
@@ -1,0 +1,177 @@
++++
+title = "Custom Roller Tutorial"
+weight = 11
+template = "doc.html"
++++
+
+This tutorial is an outline for how to set up your own L2 roller for Urbit. Some
+familiarity with how L2 works in general, as well as the role of the roller, is
+expected. See [Layer 2 Overview](/docs/azimuth/l2/layer2) for a technical
+overview of L2, [Azimuth Data Flow](/docs/azimuth/flow) to gain an understanding
+of how Azimuth data handling and processing is done on Urbit, and
+[Rollers](docs/azimuth/l2/roller) for a short summary of what the different Gall
+agents involved for rollers are. This tutorial is focused on setting up a roller
+to work on the main Ethereum network, but only minor changes are needed to
+utilize the Ropsten Ethereum network. We also explain how to set up a front end
+(Bridge) from which transactions to be batched by the roller are sent, but use
+of a front end is not mandatory.
+
+Note that this process involves giving the private key of an Ethereum wallet to
+ship running the roller so that it may spend ETH to submit transactions. If you
+do not fully understand what this entails, we recommend against running your own
+roller on the Ethereum mainnet. See [below](#step2) for more information on
+this.
+
+Henceforth, by "roller" we are referring to the ship used as the transaction
+aggregator and batch submitter, `%roller` refers to the Gall agent running on
+the roller that performs these actions, and "front-end" refers to the optional
+web GUI most users will use to interact with your roller (probably Bridge).
+
+There are three main steps involved with setting up a roller:
+ - ensuring that the data in `%azimuth` is up to date,
+ - starting and configuring `%roller`,
+ - aiming your front-end at the roller
+
+### 1. Make sure `%azimuth` state is up to date
+If you are using an ordinary live ship on the network as the roller, you should
+already have the latest `%azimuth` state and this step should not be necessary
+and you may move to step 2.
+
+If your roller is a fakezod, you will need to configure `%azimuth` to pull the
+latest [Azimuth](/docs/azimuth/azimuth-eth) data from Ethereum. The most common
+way to do this is via an Infura node, but you could instead use e.g. your own
+Ethereum node. A free tier Infura node ought to be sufficient for most people
+and is easy to set up using their tutorials.
+
+To set up your fakezod to automatically receive Azimuth updates from the Infura
+node, enter the following commands in dojo: `:azimuth %resub` followed by
+`:azimuth|watch 'https://MAINNET_INFURA_URL' %default`. Here, the url can be
+found under the Setting page for the node on infura.io listed under `ENDPOINTS`.
+
+If you do not perform this step, you'll later see an error "roller not ready"
+when the first roller batch is about to be submitted.
+
+### 2. Starting and configuring `%roller` {#step2}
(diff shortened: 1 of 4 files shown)


content/docs/azimuth/l2/roller-tutorial.md
```diff
@@ -0,0 +1,177 @@
...
+Ethereum node. A free tier Infura node ought to be sufficient for most people
+and is easy to set up using their tutorials.
+
+To set up your fakezod to automatically receive Azimuth updates from the Infura
+node, enter the following commands in dojo: `:azimuth %resub` followed by
+`:azimuth|watch 'https://MAINNET_INFURA_URL' %default`. Here, the url can be
+found under the Setting page for the node on infura.io listed under `ENDPOINTS`.
+
+If you do not perform this step, you'll later see an error "roller not ready"
+when the first roller batch is about to be submitted.
```
yosoyubik, line 52: My fault, the current behavior is that the roller-rpc agent won't accept any transactions sent via HTTP, and respond with a customer RPC error (-32003): 'Roller is not ready'

content/docs/azimuth/l2/roller-tutorial.md
```diff
@@ -0,0 +1,177 @@
...
+runtime have never been audited by a third party. We maintain a list of these
+audits [here](/audits), as well as an
+[FAQ](https://urbit.org/faq#how-secure-is-urbit) on the state of security in
+Urbit. Thus we recommend a fresh wallet generated specifically to be used by
+`%roller` and to only put the amount of ETH in it that you expect to need in the
+near future.
+
+With that being said, the command to set the private key for the wallet to be
+utilized by the roller is `:roller|setkey '0x1234567890abcdef'`, where `0x...`
+is the private key.
```
yosoyubik, line 111: We should add that it's highly recommended that the private key is only used for sending L2 batches. We have facilities for dealing with the address being used for something else, but this means that the roller will get out of sync with the expected L1 nonce, and this could cause delays in the sending batches

content/docs/azimuth/l2/roller-tutorial.md
```diff
@@ -0,0 +1,177 @@
...
+of how Azimuth data handling and processing is done on Urbit, and
+[Rollers](docs/azimuth/l2/roller) for a short summary of what the different Gall
+agents involved for rollers are. This tutorial is focused on setting up a roller
+to work on the main Ethereum network, but only minor changes are needed to
+utilize the Ropsten Ethereum network. We also explain how to set up a front end
+(Bridge) from which transactions to be batched by the roller are sent, but use
+of a front end is not mandatory.
+
+Note that this process involves giving the private key of an Ethereum wallet to
+ship running the roller so that it may spend ETH to submit transactions. If you
```
tinnus-napbus:
```suggestion
the ship running the roller so that it may spend ETH to submit transactions. If you
```

content/docs/azimuth/l2/roller-tutorial.md
```diff
@@ -0,0 +1,177 @@
...
+:roller|local  :: Configures %roller to listen to a local Ethereum node at
+port 8545. No argument.
+:roller|quota  :: Modified the number of txs a ship is allowed to send per unit
+time. Take a @ud argument.
+:roller|ropsten  :: Configure %roller to listen to a preset Ropsten Infura node.
+No argument.
+:roller|setkey  :: Load a private key into the roller and retrieves its L1
+nonce. Takes a @t argument.
+:roller|slice  :: Modified the unit of time for each ship's quota. Take a @dr argument.
+```
```
tinnus-napbus:
```suggestion
<table>
  <thead><tr><th>Dojo Command</th><th>Description</th><th>Argument</th></tr></thead>
<tbody>
  <tr>
    <td><code>:roller|commit</code></td>
    <td>Submits a new L2 batch with all pending transactions.</td>
    <td>None.</td>
  </tr>
  <tr>
    <td><code>:roller|config</code></td>
    <td>General configuration command.</td>
    <td><code>$config</code> (see <code>/sur/dice.hoon</code>)</td>
  </tr>
  <tr>
    <td><code>:roller|endpoint</code></td>
    <td>Set the Infura endpoint.</td>
    <td><code>[@t ?(%mainnet %ropsten %local)]</code></td>
  </tr>
  <tr>
    <td><code>:roller|frequency</code></td>
    <td>Sets the frequency at which batches are submitted.</td>
    <td><code>@dr</code></td>
  </tr>
  <tr>
    <td><code>:roller|local</code></td>
    <td>Configures <code>%roller</code> to listen to a local Ethereum node at port 8545.</td>
    <td>None.</td>
  </tr>
  <tr>
    <td><code>:roller|quota</code></td>
    <td>Modified the number of txs a ship is allowed to send per unit time.</td>
    <td><code>@ud</code></td>
  </tr>
  <tr>
    <td><code>:roller|ropsten</code></td>
    <td>Configure %roller to listen to a preset Ropsten Infura node.</td>
    <td>None.</td>
  </tr>
  <tr>
    <td><code>:roller|setkey</code></td>
    <td>Load a private key into the roller and retrieves its L1 nonce.</td>
    <td><code>@t</code></td>
  </tr>
  <tr>
    <td><code>:roller|slice</code></td>
    <td>Modified the unit of time for each ship's quota.</td>
    <td><code>@dr</code></td>
  </tr>
</tbody>
</table>
```

tinnus-napbus (changes requested):

content/docs/azimuth/l2/roller-tutorial.md
drbeefsupreme: @yosoyubik Sorry it has taken me so long to circle back to this.

I don't fully understand what you're saying.

> We have facilities for dealing with the address being used for something else,

By this I think you're saying that the roller can figure out the right nonce to use if the address is used for another purpose besides sending L2 transactions.

>  but this means that the roller will get out of sync with the expected L1 nonce, and this could cause delays in the sending batches

But that contradicts this. Are you saying that it can figure out the right nonce, but it will first try with the expected nonce and if that doesn't work, try the next nonce after some delay?

yosoyubik: Yeah, that second sentence is probably a bit of a worst-case-scenario so I don't think that using the L1 nonce for something else would cause delays—when the out-of-sync-nonce case is detected, after the nonce is updated, the batch is sent again right away—unless there's something else going like the private key not having enough eth... etc.

PR #1352, opened 2022-03-02
Revert "Add Slam proposal"

mikeosborne:
Reverts urbit/urbit.org#1350


mikeosborne:
I merged 1350 without seeing that there were comments. I reverted the merge not knowing it would create a new PR.

jalehman (approved):

PR #1371, opened 2022-03-11
swaps argument order in child thread example

drbeefsupreme:
would just push such a commit directly to master but I don't seem to have
permissions for that anymore


drbeefsupreme:
Added a comment about `await-thread:strandio`. This is actually currently broken (its missing an argument) and fixed with https://github.com/urbit/urbit/pull/5670 so this PR should wait on that going live.

tinnus-napbus (approved):

PR #1388, opened 2022-03-27
update rune reference docs

tinnus-napbus:
- [x] Atoms and strings
- [x] Nock . ('dot')
- [x] Wild ! ('zap')
- [x] Change Subject = ('tis')
- [x] Conditionals ? ('wut')
- [x] Cores | ('bar')
- [x] Arms + ('lus')
- [x] Cells : ('col')
- [x] Calls % ('cen')
- [x] Casts ^ ('ket')
- [x] Structures $ ('buc')
- [x] Make ; ('mic')
- [x] Hints ~ ('sig')
- [x] Imports / ('fas')
- [x] Terminators -- and ==


content/docs/hoon/reference/rune/bar.md
```diff
@@ -4,118 +4,147 @@ weight = 7
...
-be used to make wet gates that are mold builders. The buc in `|$` is a hint that
-`|$` is closely related to buc runes, and thus `|$` should be used to make mold builders, while `|*` should be used for any other sort of
-wet gate. Thus, the second argument of `|$` is frequently a buc
-rune. For further discussion of wet gates, see the entry for [`|*`](#bartar).
+`|$` is a restricted form of `|*`. The use of `|$` over `|*` is one of style, as
+either could be used to make wet gates that are mold builders. The buc in `|$`
+is a hint that `|$` is closely related to buc runes, and thus `|$` should be
+used to make mold builders, while `|*` should be used for any other sort of wet
+gate. Thus, the second argument of `|$` is frequently a buc rune. For further
+discussion of wet gates, see the entry for [`|*`](#-bartar).
```
joemfb, line 89: Another difference from `|*` is that the body of `|$` is parsed in pattern mode, to a `spec` (this is in your expansion, but might be worth explicitly noting here.

content/docs/hoon/reference/rune/bar.md
```diff
@@ -172,174 +221,246 @@ The `ne` door prints a digit in base 10, 16, 32 or 64:
...
-AST:
+`a` is a Hoon expression whose product type defines which values the gate accepts, and it usually includes a name (e.g., `n=1`). The product of `a` also serves as the default value of the sample. `b` is a Hoon expression that determines the product value of the gate.
 
-```
-[%brcn p=(unit term) q=(map term tome)]
+#### Expands to
+
+```hoon
+=>  ^~  a
+  |.  b
```
joemfb: No, it's just `=+  a  |.  b`.

content/docs/hoon/reference/rune/bar.md
```diff
@@ -553,156 +762,222 @@ with `$()`.
...
-
-`p` defines the gate sample type -- i.e., the input value type -- and usually includes a sample name assignment (e.g., `a=@`).  `q` is an expression that determines the output value of the gate.
-
-##### Expands to
+#### Expands to
 
 ```hoon
 =|  a=spec
 |%  ++  $  b=hoon
 --
```
joemfb, line 831: The `^~` is actually here, so it'd be `=+  ^~(*a)  |% ...`. Not sure if you want to go to that level of detail ...

tinnus-napbus: is there an easy way to figure out these expansions? I've just left these as they were cos they were so painful to validate. Not sure these are even useful to people tbh

tinnus-napbus: I would prefer to strip them cos for 99% of people they'll be more confusing than illuminating but maybe it's desirable to have them

tinnus-napbus: I tries `~(open ap (ream '(add 1 1)'))`, is that how you'd do it? Is there a way to get something more readable than straight AST?

joemfb: `+open:ap` expands the outermost hoon it's given, incrementally. Synthetic runes are expanded in the compiler as they're encountered -- there's no function to fully expand. And even if there were, it'd probably be too much detail -- these sections are an editorialization on the actual macro-expansions. I think it's very important to understand the natural/synthetic distinction, and the general shape of the macro-expansions ("`|=` is just `|_` is just `=+` and `|%`"). But I don't know if the expansions in the docs are the best way to convey that. It might be enough to have a tutorial on `+ream`, `+open:ap`, `+mint:ut`, &c, and just note whether a rune is natural or synthetic.

As for something more readable than the raw AST, you could try `+render-hoon` in `lib/pprint` (but idk how good that output will be -- I've seen it spit out a bunch of "gensyms" ...).

tinnus-napbus:
ok I've changed the expansions you specified (and addressed other comments), idk if there are any other expansions which are egregiously wrong

tinnus-napbus:
@sigilante I've restructured the rune docs, updated outdated/incorrect info, added missing runes, clarified some obtuse explanations, and added a new fas section. They're approximately structured in line with how you suggested, though they don't all have a semantics sections, I've mostly retained the existing explanatory section names to avoid more significant rewriting

content/docs/hoon/reference/rune/sig.md
```diff
@@ -314,139 +591,277 @@ From the kernel:
...
-##### Product
+#### Product
 
 Pretty-prints `p` on the console before computing `q`.
 
-##### Discussion
+#### Discussion
 
-This rune has no semantic effect beyond the Hoon expression `q`.  It's used solely to create a side-effect: printing the value of `p` to the console.
+This rune has no semantic effect beyond the Hoon expression `q`. It's used
```
sigilante, line 645: There's an issue with type matching e.g. for cords.  So we should recommend using tapes or figure out exactly why that happens and note it here.

content/docs/hoon/reference/rune/tis.md
```diff
@@ -4,127 +4,286 @@ weight = 5
...
+  --
+  (foo 42 27)
+[27 42]
+```
 
+---
+
+## `=|` "tisbar"
+
+Combine a default type value ("bunt value") with the subject.
```
sigilante: This phrasing seems confusing to new learners... let's pattern it after =/ and =+ which “combine a named noun with the subject”:

“Combine a named noun with the subject; assign it the bunt value” or similar

content/docs/hoon/reference/rune/lus.md
```diff
@@ -4,33 +4,50 @@ weight = 8
...
-The `+|` doesn't produce an arm.  It instead provides a label for the arms that follow it.  The arms of a core can be divided into **chapters** for 'organization'.  Chapter labels aren't part of the underlying noun of the core; they're stored as type system metadata only.
+The `+|` doesn't produce an arm. It instead provides a label for the arms that
+follow it. The arms of a core can be divided into **chapters** for
+'organization'. Chapter labels aren't part of the underlying noun of the core;
+they're stored as type system metadata only.
 
 See `tome` in the Hoon standard library.
 
-##### Examples
+#### Examples
```
sigilante, line 53: Let's put a pin to work on these.  These examples are distinctly unhelpful to someone deciding whether/how to use +|.

content/docs/hoon/reference/rune/sig.md
tinnus-napbus: I'm not aware of a type issue with `~&`

tinnus-napbus: oh you mean how if eg you return either a `@ud` or `@t` it always prints an `@t`. This isn't to do with `~&`, it's because any kind of conditional fork expression eg:

```
?:  ...
  12
 'foo'
 ```
 
 is evaluated to return a union of its possible types, in this case `?(@ud @t)`. The problem is that `$?` can't differentiate between auras so it just casts everything to the default which is the last item in the union, a `@t` in this case.
 
 so the problem isn't `~&`, the problem is the limitations of the type system. You'd have the same problem in any case where you were returning a union of auras

content/docs/hoon/reference/rune/lus.md
tinnus-napbus: lusbar is basically completely useless. You can't reference chapter labels short of manually processing the $type of a core. I can't think of a better example cos there's really isn't a practical example. I've added a couple of notes about how you wouldn't use this

PR #1405, opened 2022-04-11
threads: use correct start arguments

Fang-:
Generally, you want to use the latest case when starting threads.
The agent's build case may not be latest, and clay may not support building threads from there.

May or may not need similar changes elsewhere.

cc @yosoyubik, who ran into this.


tinnus-napbus:
ohhh yeah the beak contains the build case, I was thinking its case would be now but that maeks sense

tinnus-napbus (approved):

PR #1412, opened 2022-04-19
build: upgrade to webpack 5, next 12

matildepark:
Just the usual "keeping things current" push before we do larger work ...

- Upgrades us to next.js 12
- next.js 11 moved to webpack 5, so this part of the README is removed.
- Pins the site build to node 14, gives instructions on using `nvm` in README.
- Upgrades the site from Tailwind 2 to Tailwind 3


matildepark:
Ooh, please hold, preview is showing weirdness (seemingly with the globals css). Matomo also coming up undefined, but I know very little about the deployment there (perhaps development previews don't have it in the environment anyway).

matildepark:
That looks better now.

matildepark:
Stepped through the entire site in preview and feel satisfied that we can rip the bandaid off.

PR #1413, opened 2022-04-19
packages: bump remark-gfm to 3.0.0

matildepark:
Notably, **makes footnotes work**! It was driving me crazy already (see @sigilante's latest post, for clearest example, in the preview).

[No breaking changes I can make of](https://github.com/remarkjs/remark-gfm/releases), though perhaps we can remove it from transpilation. Unclear.

Diff size due to package-lock version update — node 14 and 16 alike both use a newer version. I'll bump us to node 14 in #1412.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,8 +1,8739 @@
 {
   "name": "urbit.org-2",
   "version": "0.1.0",
-  "lockfileVersion": 1,
+  "lockfileVersion": 2,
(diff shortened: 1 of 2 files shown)


sigilante:
LGTM as long as it isn't breaking.  No breaking changes are flagged in their source repo.

sigilante (approved):

PR #1417, opened 2022-04-22
search: debounce input

### LINKED ISSUE #1330: search is swallowing stdlib glossary entries again
this was fixed for a while but it's started happening again. When you search for e.g. `turn`, it flashes the `/lib/glossary.js` result for a second and then it disappears. If you backspace the last letter and retype it, the glossary result then correctly appears.

matildepark:
We weren't debouncing search input, so we would send queries to the back-end each keystroke; sometimes partial keystroke requests would return after our complete one.

Fixes #1330

@tinnus-napbus can you run the deploy preview here and do your darndest to replicate your scenario again? I think we are good.

### DIFF
--- a/components/Search.js
+++ b/components/Search.js
@@ -1,6 +1,7 @@
 import { Component, createRef } from "react";
 import { glossary } from "../lib/glossary";
 import { withRouter } from "next/router";
+import debounce from "lodash.debounce";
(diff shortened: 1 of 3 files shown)


tinnus-napbus (approved):
yup seems to be working correctly now

PR #1421, opened 2022-04-26
Updating to account for settlement in stars

thelifeandtimes:
movement away from $WSTR means that as of yet uncontracted deals should be priced in stars. @drunkplato would be good to get your eyes on this to make sure it feels right. could push a star from milestone 2 into milestone 3 if that feels better?


jalehman (approved):

drunkplato:
Looks good to me. Feel free to merge.

PR #1427, opened 2022-04-28
Replace getting started guide and add planet file

joelhans:
For some context: This is a project I've been working on with @jalehman/~wolref-podlex "out of band," and we agreed it was time to bring this into GitHub for a more comprehensive review and do any preparation to make sure a change of this magnitude goes through smoothly.

The overall goal is to improve the onboarding experience by offering users a few "chosen" paths toward getting on the network:

- Easy = Port + Comet
- Fun = VPS + CLI + Planet
- Hosted

And from there, walk them through the basics of adding groups and installing apps so that they have all the information they need to start exploring.

A few notes and TODOs:

- I had originally imagined that by clicking one of the **Easy**/**Fun**/**Hosted** boxes, users would see *only* those specific steps, with a **back** button to let them choose different options. I'm not 100% sure whether that's a better experience than what we have right now. I'd probably need some development help to make that selector experience possible if that's the way we go.
- I still have to add a few images—they should all be marked in the document.
- There's this comment to address from @gordoncc, which I'm still thinking about:
  > While I like the segmentation here, I'm wondering if we could improve the headings.  It's a bit confusing when skimming the document to see "Easy" and "Fun" - sections -- they seem like sequential steps rather than separate paths.  While a bit boring maybe "Local / Cloud / Hosted" would be more clear

### DIFF
--- a/content/getting-started/_index.md
+++ b/content/getting-started/_index.md
@@ -1,27 +1,106 @@
 +++
-title = "Getting Started"
+title = "Getting started with Urbit"
 weight = 0
 description = "How to install Urbit."
 page_template = "page_indiced.html"
 template = "getting-started/index.html"
 +++
 
-Don't know what Urbit is? Spend ten minutes or so reading our page on
-[Understanding Urbit](/understanding-urbit/).
+Nothing about building a new OS from scratch has been easy. Or creating an identity system that adequately replaces the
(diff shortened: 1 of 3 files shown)


matildepark:
Superseded by #1488.

PR #1435, opened 2022-05-02
Create canvas.md

gordoncc:
Adding canvas app public link page

### DIFF
--- a/content/applications/~dister-norsyr-torryn/canvas.md
+++ b/content/applications/~dister-norsyr-torryn/canvas.md
@@ -3,7 +3,7 @@ title = "canvas"
 shortcode = "~dister-norsyr-torryn/canvas"
 license = "MIT"
 image = "https://yosoyubik.fra1.digitaloceanspaces.com/canvas/tile.svg"
-bgColor = "#3787B9
+bgColor = "#3787B9"
 developer = "~dister-norsyr-torryn"
 website = "https://github.com/yosoyubik/canvas"
 +++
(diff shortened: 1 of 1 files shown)


gordoncc:
npm is unhappy for some reason

gordoncc:
> npm is unhappy for some reason

npm very unhappy with me not closing the bgcolor hex value

content/applications/~dister-norsyr-torryn/canvas.md
```diff
@@ -0,0 +1,12 @@
++++
+title = "canvas"
+shortcode = "~dister-norsyr-torryn/canvas"
+license = "MIT"
+image = "https://yosoyubik.fra1.digitaloceanspaces.com/canvas/tile.svg"
+bgColor = "#3787B9"
+developer = "~dister-norsyr-torryn"
```
matildepark: The dev is `~norsyr-torryn` actually

gordoncc: i've been conflating "developer" and "software provider" -- need to make a distinction between the two more clear and update some of the app pages and install guide

matildepark: The provider is in the shortcode; developer should be an entity or a patp right? Many ppl distribute via moons but it makes it hard to assess original author if both fields are the moon.

gordoncc: Yes, the developer _should_ be an entity or patp.  However, in Landscape they use the terms "developer" to indicate an urbit address that could be a moon, star, anything, and "developer desk" to indicate the full shortcode.  So if we say "Tlon" is the developer then it's very confusing when "dozzod^dozzod" is also the developer.

gordoncc: Chatting with James at Tlon now to see how permanent that nomenclature is

PR #1440, opened 2022-05-02
redirect from takeover to week

jalehman:
Ideally the URL of the group should match the name of the group. This should be fine?


matildepark:
I agree, it should. This will redirect to a 404 though, you need to actually rename the file too.

jalehman:
I did, right? https://github.com/urbit/urbit.org/blob/dcec828da50302fdf59493677fccaab58e588d2f/content/groups/~ladtem-filmyr/urbit-ny-week.md

matildepark:
The file says ny-week.md and the redirect is to nyc-week.md. Which one is canonical?

jalehman:
Alright, third try's the charm.

matildepark (approved):

PR #1443, opened 2022-05-03
markdown.css: preserve whitespace in inline code blocks

### LINKED ISSUE #1332: whitespace issue: inline code is rendering gaps (`  `) as aces (` `)
E.g. `:-  1  2` is rendering as `:- 1 2`, see near the bottom of https://urbit.org/docs/hoon/hoon-school/hoon-syntax for an example, but it seems to be happening all over the site.

Not sure when this happened. I tried looking at vercel deployments on PR's from 6+ months ago and it was still rendering that way, and I know it hasn't been broken for that long, but I also don't know the details behind how those are generated.

matildepark:
Going surgical with the CSS here — does appear to remedy the issue as described in original ticket.

image

Check the preview deployment and compare, reviewers!

Fixes #1332

### DIFF
--- a/styles/markdown.css
+++ b/styles/markdown.css
@@ -73,6 +73,10 @@
   @apply text-base leading-relaxed;
 }
 
+.markdown.technical p code {
+  white-space: pre-wrap;
+}
+
 .markdown.technical li {
   @apply py-2 ml-6 text-base;
 }


tinnus-napbus (approved):
nice

PR #1445, opened 2022-05-03
TableOfContents: smaller fonts

### LINKED ISSUE #1135: wrapped section titles in contents sidebar indistinct
section headings which are too wide get wrapped in the auto-generated contents sidebar thing, but their line spacing/padding/whatever is identical to the spacing between separate sections. This makes it unclear at a glance that the two lines are a single thing as opposed to two separate sections.

Eg:

2021-09-18_03h19m41s

It would be cool if the line spacing or whatever were smaller (or otherwise visually distinct) for wrapped section headings so it's clear they're one thing.

@g-a-v-i-n 

Edit: These pages have live examples:

https://urbit.org/docs/arvo/tutorials/move-trace
https://urbit.org/docs/arvo/eyre/examples

matildepark:
Fixes #1135 

Title uses regular font size; headers use `text-sm`, subheaders use `text-xs`. Lets us fit more on a line but also gives a better sense of content hierarchy beyond just colours.

Compare (before / after):

Screen Shot 2022-05-02 at 5 32 38 PM
Screen Shot 2022-05-02 at 5 32 22 PM

See `docs/arvo/tutorials/move-trace` on the deploy preview for example and the original issue has current links for comparison.

### DIFF
--- a/components/TableOfContents.js
+++ b/components/TableOfContents.js
@@ -55,13 +55,13 @@ const Headings = ({ headings, activeId }) => (
     {headings.map((heading, index) => (
       <li key={heading.id}>
         <a
-          className={index === 0 ? "font-bold" : "font-medium"}
+          className={index === 0 ? "font-bold" : "font-medium text-sm"}
           href={`#${heading.id}`}
         >
           {heading.title}
         </a>
         {heading.items.length > 0 && (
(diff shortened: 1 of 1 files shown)


gordoncc (approved):
Very nice typographic adjustment

PR #1446, opened 2022-05-03
docs: add "edit page on github" link

### LINKED ISSUE #790: docs: add the ability to submit a PR against the docs on every single page
Within Tailwind's css docs, you can see at the bottom of their pages a link to [submit a PR](https://tailwindcss.com/docs/background-origin#disabling) against the page the link was followed from.

We often run into intrepid urbit devs who run into various frustrations w/ our current docs, and establishing a means of quickly submitting patches to the docs could help us out in a time when we're trying to grow more urbit devs.

matildepark:
Fixes #790

Current implementation has it at the bottom, but it seems too loud. Where should this go and how should it be presented?

image

### DIFF
--- a/components/ContentArea.js
+++ b/components/ContentArea.js
@@ -44,7 +44,7 @@ export default function ContentArea(props) {
     <div className="w-full min-w-0 flex flex-col items-center">
       <div
         ref={scrollBox}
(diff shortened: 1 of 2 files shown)


gordoncc:
Old designs have it like this in the mocks --- looks good to me.  I'd add in a green hover state.
Screen Shot 2022-05-04 at 4 53 01 PM

matildepark:
Amended. I tweaked ContentArea so the bottom has less padding since pages kept ending in a really awkward position. In general I feel like docs / getting started layout could use a redesign; it looks odd to my eye.

gordoncc (approved):

PR #1470, opened 2022-05-04
grants: tweaks as per @mikeosborne

matildepark:
- Displays champions and mentors
- Grantee, champion, mentor all clickable
- Exposes grant IDs


mikeosborne (approved):
Beautiful!

PR #1473, opened 2022-05-05
build: refactor buildPageTree into a prebuild script

matildepark:
Part 1 of some performance work.

`buildPageTree` iterates through a directory and spits out a recursive tree for sidebars. Currently on build, we run it *twice* **per page** which adds up in large directories like docs.

This refactors the `buildPageTree` script as a CommonJS script that takes `directory` and `sort` arguments and outputs a `[dir].json` object, and runs it against a few directories as a prebuild script at the start of development or build.

Build time before/after:
- **Docs**: 182959ms -> 160406 ms
- **Updates**: 30562ms -> 25925ms

Etc etc.


matildepark:
All in all the deploy goes from 6m 17s to 5m 54s. Everything works, merging ...

PR #1483, opened 2022-05-06
Create desire-lines.md

gordoncc:


matildepark:
Heads up that most of these links need a space after the preceding period and the group mentioned at the end could be a group link.

gordoncc:
> Heads up that most of these links need a space after the preceding period and the group mentioned at the end could be a group link.

TY!

matildepark (approved):

PR #1485, opened 2022-05-06
grants: make query assignment reactive

### LINKED ISSUE #1477: grants page doesn't properly filter by grant type on refresh
@matildepark, thinking this might be a quick fix for you.

When filtering by the different grant types on urbit.org/grants, selecting the various grant type buttons (proposal, apprenticeship, bounty) will properly filter the visible cards, and will update the url utms, but if you refresh the page or otherwise share the url in the interest of sharing a specific set of grants to an outside party (i.e. 'open apprenticeships' using https://urbit.org/grants?program=apprenticeship&open=true&completed=false&wip=false), it will show all of the grant types. Seems like the other tags work correctly though.

matildepark:
Fixes https://github.com/urbit/urbit.org/issues/1477. State in the grants interface could use a refactor (it comes from a few different sources and could probably have its logic unified), but this PR tries to structure assignment for our interface to be more reactive to query parameters.

@thelifeandtimes Test this out in the deploy preview for me? Works in my environment.


thelifeandtimes:
@matildepark this still doesn't work for me. try this link and see what you see: https://urbit-8meybt8v5-urbit.vercel.app/grants?program=apprenticeship

matildepark:
> @matildepark this still doesn't work for me. try this link and see what you see: https://urbit-8meybt8v5-urbit.vercel.app/grants?program=apprenticeship

Ah, that's a different thing entirely. I fixed it for "Complete vs. In Progress" etc etc. Those aren't queries right now at all I think? So they'd have to match that state management.

matildepark:
@thelifeandtimes @jalehman I completely rewrote the logic on the Grants page to exclusively use queries for state management instead of only partially using them; please check the branch to double check my implementation maintains the logic you expect — it appears to be the same to me in testing. I also made it so that permalinks you make when filtering the grants down jump down to the view grants section.

thelifeandtimes (approved):

PR #1487, opened 2022-05-07
/docs, /understanding-urbit, /using: show current page in sidebars as green

matildepark:
Another thing we lost around October — sidebar selected items used to be green. Now they are again.

image

Also tweaking the dot CSS a little to match the disparate situations it's used in — on Getting Started it's used for header items, which looked a little awkward, and it always sort of looked uncentred.


gordoncc (approved):
Lookin good

PR #1488, opened 2022-05-07
feat: new getting started guide

matildepark:
Supersedes #1427. Implements the aforementioned PR using @gordoncc design mockups. Adds descriptions to the docs templates.


gordoncc:
- [x] Cloud hosting guide link on "Fun" page is broken, should point to "/using/running/hosting"
- [x] Re-organize left hand navigation per attached image
- [x] Move "Layer 2 for planets" page to User's Manual / Urbit ID
- [x] Update "Get a planet" page with new content / elements from new page in Figma
Screen Shot 2022-05-09 at 10 23 24 AM

tacryt-socryp:
Note: I can't figure out how to even find the getting started guide from the homepage.

matildepark:
> Note: I can't figure out how to even find the getting started guide from the homepage.

There's a big green "Get Started" link at the top? I'm assuming you're looking at mobile, where it's missing for some reason — presumably it was not included because you can't actually get started from a phone.

matildepark:
My input wasn't asked for but personally I would be very frustrated trying to find the CLI instructions — a lot of people just come here to set up Urbit again somewhere else and these "total newbie" instructions wall off that user story entirely.

gordoncc:
> My input wasn't asked for but personally I would be very frustrated trying to find the CLI instructions — a lot of people just come here to set up Urbit again somewhere else and these "total newbie" instructions wall off that user story entirely.

I'm not averse to keeping the CLI Instructions in the sidebar under an advanced subhead for now
Screen Shot 2022-05-09 at 2 57 08 PM

gordoncc:
Ok I think we're getting there.  

I like subheadings on the left, and think we should roll with that and rename the sections "Getting Started" and "Additional Guides."  In there we can move the guides for joining groups and installing apps that we did for the public link pages.  (Let's update any links or setup a redirect from the /guides/ pages)

The Easy/Fun/Hosted nomenclature really doesn't feel right to me, especially on the sidebar.  I find it confusing.  So let's rename the sections per the mocks below (see Figma for updated subheadings on the pages.)

Finally, I updated the layout and copy of the "Launch into Urbit" section for greater legibility.

Screen Shot 2022-05-13 at 11 09 15 AM
Screen Shot 2022-05-13 at 11 24 00 AM
.

matildepark:
If subheadings are going to be a pattern going forward I'm going to need to do some rearchitecture instead of hard coding it for this one time.

matildepark:
Also I don't really understand "not a permanent way to use Urbit" — you can certainly use Port practically indefinitely. It has tradeoffs to consider regarding mobile access, hosting groups, general "uptime", but so does stuff like IPFS or really any P2P app you run on your computer, and there are guides working around those kinds of issues if necessary. If you casually visit the network you don't really have to worry about anything.

matildepark:
Batter up @gordoncc

tacryt-socryp:
Still think it's bad and wrong that this isn't discoverable from mobile? People might at least want to be able to find and read it from the homepage.

jalehman:
The lack of mobile accessibility wasn’t introduced in this PR. Perhaps it can be worked in, but it’s not the subject of what we’re trying to do here.

—

~wolref-podlex
https://urbit.org ( https://urbit.org/ )

matildepark:
I did reply over email, but it didn't make it here for some reason ... 

Basically what @jalehman said; if @gordoncc has a mock for how he thinks mobile flow works (just put it in the sidebar? Etc etc) I can do it. Ideally we don't just send them to a page they can do nothing on; I think that's why it wasn't put in when it was first designed.

gordoncc:
Yeah not having a link to Getting Started, Manual, or FAQ on the mobile menu just seems like an oversight here.  @matildepark let's just throw them in when you get a chance.

I went ahead and pushed a bunch of commits, I think the guide is looking good to deploy now.

gordoncc (approved):
A great improvement from what's on the site now, let's get after the mobile menu navigation next!

matildepark:
My point was more than it wasn’t in scope for the deploy. If Gordon has a
mock for how he wants the mobile flow to go I’m down.

PR #1514, opened 2022-05-18
update apprenticeship comp

thelifeandtimes:
Would appreciate a review+merge.


mikeosborne:
Are we choosing not to specify mentor compensation?

thelifeandtimes:
@mikeosborne yea, for now. I think we'll evolve towards the right answer on this one, but because we don't want to muddy the water on the wstr relationship ship I think we keep it off the clear web for the time being.

mikeosborne (approved):

PR #1518, opened 2022-05-18
add eyre-scry.md

ixv:
cc @liam-fitzgerald @belisarius222

### DIFF
--- a/content/grants/eyre-scry.md
+++ b/content/grants/eyre-scry.md
@@ -1,5 +1,5 @@
 +++
-title = "Eyre Request Scry Cache"
+title = "Eyre Request Cache"
 date = "2022-05-18"
 
 [taxonomies]
@@ -8,13 +8,16 @@ grant_category = [ "Core Dev" ]
 
 [extra]
 image = ""
-description = "Speed up Urbit http server by implementing a scry mechanism and cache for requests"
-reward = "10 stars"
+description = "Speed up Urbit http server by implementing a cache for requests"
+reward = "5 stars"
 assignee = ""
+champion = "~tirrel"
+grant_id = "B0182"
 completed = false
 canceled = false
-link = ""
+link = "https://airtable.com/shr4qt9t9kz7RaOIa?prefillGrant+ID=&prefillGrant+Name=Eyre%20Request%20Cache"
 +++
+
 
 ## Rationale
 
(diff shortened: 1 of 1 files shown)


Fang-:
Notable prior art available here: urbit/urbit#4674

thelifeandtimes (commented):
@ixv A few comments from the perspective of the foundation's grant program:
- Is this a proposal intended to be taken on by the ~tirrel team as a whole? assignee field is blank. Or is it a proposed bounty that you guys are speccing out and would like to see someone else take on and the foundation to help fund?
- Max proposal size at present is 5 stars, and ideally those map to milestones. I understand this may be a slightly different undertaking, but if the Eyre and Vere milestones are happening sequentially we should split them into two different proposals with more narrowly defined milestones. 
- If this is a proposal (as noted in the taxonomies) from ~tirrel, what would be the proposed delivery timeline?
- assuming notes under the implementation header are forthcoming?

ixv:
- We're speccing it out and hoping that someone else will take it on who we will manage. Funding will come from our stipend as per our arrangement.

- I'm not totally opposed to splitting this into 2 proposals since they are atomic pieces of work that can go in separately, but given that the 2 parts of work are pretty closely intertwined in that the APIs for both are pretty much the same, it might be too much of a coordination burden.

- I would estimate 6 months total, or 3 months for each milestone. Will update the PR shortly.

- The implementation header is meant to include the milestones, ie: those are the 2 things that need to be implemented. I feel they have sufficient detail, but if you disagree I can probably write up some more.

thelifeandtimes:
Ah, noted on the implementation header. Let's split into 2 grants so that we can line up the correct person and there is an additional 'carrot' for successful execution of phase 1 that they can then go take on phase 2, or if things aren't fully working out we can find the next contributor and have it be less messy. I'll also change to a bounty so we can list and start getting applicants (vs a proposal which would already have a specific contributor attached to it).

thelifeandtimes:
@ixv @tacryt-socryp you guys going to split this into 2?

thelifeandtimes:
@matildepark, know why this is breaking?

matildepark:
> @matildepark, know why this is breaking?

Looks like it built to me.

mikeosborne (approved):

PR #1523, opened 2022-05-24
search: prioritise exact matches

matildepark:
@ynx0 Fixes in my testing, can you try some similar terms?

### DIFF
--- a/components/Search.js
+++ b/components/Search.js
@@ -24,11 +24,26 @@ class Search extends Component {
   }
 
   glossarySearch(query) {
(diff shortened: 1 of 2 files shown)


ynx0:
So `cat` and `snag` for instance work now but, `dif` and `tap` (for the `by` core) don't work. Same for partials of exact matches, such as `ala` for `alas` or `lim` for `limb`, `tim` for `time`, etc.

matildepark:
Closing in favour of #1528.

PR #1540, opened 2022-06-02
Markdown migration to Markdoc

### LINKED ISSUE #1299: RFC: Add functionality to inline duplicate content across markdown files
In a recent PR (#1295), I fixed a spelling bug between two sections with identical contents replicated manually on two separate pages. Now, this wasn't really a big deal, but I feel like these kinds of mistakes, along with the hassle of constantly updating the relevant sections, could be mitigated by some kind of "markdown include" functionality. However, a cursory search about including markdown files from other markdown files doesn't seem to be supported in nextjs. I did stumble upon [Paradox](https://developer.lightbend.com/docs/paradox/current/directives/includes.html), which has this kind of functionality and quite frankly a lot more, but migrating the whole documentation system would require a lot more impetus than just this. I'd love to hear thoughts about how this could be resolved, if it seems like a reasonable enough issue.

<br/>

Pages with duplicate content in question:
- https://urbit.org/docs/userspace/gall-guide
- https://urbit.org/docs/userspace/gall-guide/intro

### LINKED ISSUE #1358: HTML gets double-parsed
either just HTML or all markdown in general seems to get double-parsed. 

For example, in [this doc](https://raw.githubusercontent.com/urbit/urbit.org/master/content/docs/hoon/guides/sail.md) I had to convert HTML entities into HTML entities, so `&lt;` is `&#x26;&#x6C;&#x74;&#x3B;`, in order for it to render correctly. Additionally, HTML gets parsed as HTML even when it's in fenced codeblocks.

### LINKED ISSUE #1471: performance improvements
Our builds take longer than 5 minutes; right now it's fine if you're developing on the site actively — `npm run dev` only generates content as it needs it, but if you're constantly merging PRs then 5 minutes per build adds up.

- [x] Pull repetitive directory construction logic into a centralised place. The docs pages run `buildPageTree` twice **per page** each build, iterating through the full content/docs directory to construct the sidebar. I have some work stashed here to pull that into a one-time pre-build script that just dumps the JSON for reference, but it only got us a 7.9% performance increase.
- [ ] Markdown rendering takes nearly all of the time here, because we're doing it server-side on build into HTML instead of client-side Markdown rendering. It's in our interests to move the Markdown pipeline back to client side and assess how to migrate our syntax highlighting in that environment — if prism.js exports ESM now, we might be fine.

matildepark:
- Markdown build pipeline moved to client-side rendering (fixes #1471, fixes #1358). Builds finish in 2.5 minutes, as opposed to 5+.
- Migrates the entire build pipeline over to [Markdoc](https://markdoc.io), allowing for construction of React components in-markdown, transclusion of common markdown content, and incorporating functions inside the template language. Fixes #1299

Markdoc doesn't like it when you use its template format inside code blocks, which we do — see https://github.com/markdoc/markdoc/issues/72. It was producing builds that would tank my computer, eating all my RAM, and I had to manually delete folders and escape Markdown components one section of the site at a time to figure out the cause. We escape them a little in 20a2135ef5d4ea151cfd622b2ffdb03fcb6e9cdf with apologies to @tinnus-napbus.

We can't just inject Markdown plugins with Markdoc so I rewrote (or ported from other repos) all the features the old Markdown pipeline gave us but still some left to catch and some quirks still here —

- ~Markdoc doesn't support inline HTML (https://github.com/markdoc/markdoc/issues/10). I had to parse through every piece of inline HTML on the site and render it as Markdown. The getting started pages use a Tab component now instead of the hard-coded selector items.~
- ~I still have to make external links open in a new tab and make images slightly oversized.~

I patched our Markdoc dependency for the remaining features, see [381e97b](https://github.com/urbit/urbit.org/pull/1540/commits/381e97b693db587b2ea33793ba6b7d44473cef62).

It's not clear what's going to be quirky in the build so while I come back around to finish the little pieces here, I'm leaving this here as a draft for people to peruse the site deployment with.


matildepark:
- Brought our links and images to the same style / "opens in new tab" behaviour
- Our tables were written in inline HTML so I manually rewrote them in Markdoc format
- We however use inline HTML all over the place for superscript, iframes, inexplicably some blog posts are just written in HTML ... I could port *most* of this by hand but the iframes and movies wouldn't make it. I'm gonna see how to implement the inline HTML feature for the remainder, then this should be okay to ship.

matildepark:
Patched dependencies to support HTML content and superscript; migrated superscript from `<sup>` tags to markdown; migrated our inline `<div>`s from the past few features and guides to appropriate components.

Inexplicably the deploy built in like a minute. That's ... weirdly fast, but the deploy looks right. I think this is ready for review if either of you want to peruse the deploy.

matildepark:
So today I picked out a visual regression test script ([this one](https://github.com/nalmeida/master-of-puppets)) and compared our production deploy to the Vercel deploy; screenshotting and comparing them made a new coloured diff, and I went through them. Basically this has slight differences in padding/spacing for code blocks, but the site by and large looks the same. Some quirks in Markdown processing got corrected in the last commits. Our artisanal footnote parser trips up on stuff with tons of footnotes, but we only have a single post like that, so I tweaked it. I feel about ready to rip off the bandaid.

PR #1545, opened 2022-06-05
Add vere reference

tinnus-napbus:
Adds a reference page for the `urbit` binary in the user manual. Details all the options & utilities, explains its basic features, and gives examples of common usage


content/using/running/vere.md
```diff
@@ -0,0 +1,632 @@
...
+  `x86_64-darwin`, and `x86_64-windows`, though more may be added in the future.
+- `-v, --version VER` - version number, `VER` is e.g. `1.9`.
+- `-p, --pace` - release channel, e.g. `live`.
+
+Example usage:
+
+- Undocked: `urbit vere -a x86_64-linux -v 1.9 -p live .`
+- Docked: `[pier]/.run vere -a x86_64-linux -v 1.9 -p live .`
+
+### `serf ARGS`
```
sigilante, line 301: At some point in this doc want an aside about what king v. serf are responsible for.

content/using/running/vere.md
```diff
@@ -0,0 +1,632 @@
...
+Replay up to the specified event `NUMBER`.
+
+- Undocked: `urbit -n 10000 [pier]`
+- Docked: `[pier]/.run -n 10000`
+
+A snapshot will be saved in a jamfile at `[pier]/.urb/roc/[NUMBER].jam`.
+
+### `--no-conn`
+
+Disable the control pane unix socket.
```
sigilante, line 484: Main case so far is ARM/RPi seems to not support.

content/using/running/vere.md
```diff
@@ -0,0 +1,632 @@
...
+lmdb: 0.9.29
+curl: 7.81.0
+```
+
+- Undocked: `urbit -R [pier]`
+- Docked: `[pier]/.run -R`
+
+### `-S, --skip-battery-hashes`
+
+Disable battery hashing. This option is currently a no-op, and doesn't do
```
sigilante: ??? as an aside for me, what was this?

sigilante (approved):
Looks great to me.

content/using/running/vere.md
tinnus-napbus: my ships run on arm and it appears to work fine, didn't need to disable

content/using/running/vere.md
tinnus-napbus: ¯\\_(ツ)_/¯ @joemfb do you know?

content/using/running/vere.md
tinnus-napbus: I have added an explanation into the second paragraph at the top (king = I/O & event persistence, serf = computation & state persistence)

content/using/running/vere.md
```diff
@@ -0,0 +1,633 @@
...
+use a lot of memory to complete, sometimes as much as 8GB, so if you don't have
+ample memory on your machine you might need to map some swap space.
+
+To run `meld`, you can either do `urbit meld /path/to/pier` or
+`/path/to/pier/.run meld` if it's been docked.
+
+If the meld succeeds, it'll print a memory report and exit. If it exits with
+`KILLED`, it means it ran out of memory and the meld was aborted.
+
+**Note that the ships must not be running when using either of these.**
```
Fang-: Consider moving this up to _before_ the mention of the commands to do this.

joemfb: All stateful subcommands take the pier lock as of v1.9, so this is no longer a safety issue. It's probably a good note regardless, with lots of old binaries out there.

content/using/running/vere.md
```diff
@@ -0,0 +1,633 @@
...
+
+- `-a, --arch ARCH` - architecture, `ARCH` may be one of `x86_64-linux`,
+  `x86_64-darwin`, and `x86_64-windows`, though more may be added in the future.
+- `-v, --version VER` - version number, `VER` is e.g. `1.9`.
+- `-p, --pace` - release channel, e.g. `live`.
+
+Example usage:
+
+- Undocked: `urbit vere -a x86_64-linux -v 1.9 -p live .`
+- Docked: `[pier]/.run vere -a x86_64-linux -v 1.9 -p live .`
```
joemfb: the `vere` doesn't really support docking -- it's always the same command (the pier is irrelevant). It's intended for semi-manual upgrades: `urbit prep $pier && urbit vere .`, or something like that.

content/using/running/vere.md
```diff
@@ -0,0 +1,633 @@
...
+
+### `serf ARGS`
+
+Run as a 'serf' (as though it were the previously separate `urbit-worker`).
+
+- Undocked: `urbit serf [pier] [key] [flags] [cache-size] [at-event]`
```
joemfb, line 308: I'd note that this is for internal IPC -- it's a programmatic interface, not a user facing CLI.

content/using/running/vere.md
```diff
@@ -0,0 +1,633 @@
...
+
+- Undocked: `urbit -w sampel-palnet -G 0w50w.642KA.glYh8.RvaRp.6TA35.3XRYN.ULCI6.UmcsZ.5rGvG.3dvFt.E6rMF.mZnd~.3O6QB.7LZ1f.-ukYN.z2fcY.Ecviv.LiZEg.8w0sp`
+
+### `--http-port PORT`
+
+Set the HTTP port to `PORT`. This is the port you'd insecurely access Landscape
+on. Note this is for insecure HTTP only, see `--https-port` for HTTPS.
+
+- Undocked: `urbit --https-port 8085 [pier]`
+- Docked: `[pier]/.run --https-port 8085`
```
joemfb: Typo, s/b `http-port` here.

content/using/running/vere.md
```diff
@@ -0,0 +1,633 @@
...
+### `--no-conn`
+
+Disable the control pane unix socket.
+
+- Undocked: `urbit --no-conn [pier]`
+- Docked: `[pier]/.run --no-conn`
+
+### `--no-dock`
+
+From v1.8 onwards, when a new ship is booted, the urbit binary will copy itself
```
joemfb: `v1.9` onwards.

content/using/running/vere.md
```diff
@@ -0,0 +1,633 @@
...
+- Docked: `[pier]/.run -e https://the.url`
+
+### `-F, --fake SHIP`
+
+Boot a new fake development ship. Real networking is disabled, and instead the
+fake ship can only communicate with other fake ships on the local machine.
+
+```
+urbit -F zod
+```
```
joemfb, line 386: This works for all ship classes except comets.

content/using/running/vere.md
```diff
@@ -0,0 +1,633 @@
...
+
+### `-g, --gc`
+
+Set GC flag, dumping garbage collector logs to stdout.
+
+- Undocked: `urbit -g [pier]`
+- Docked: `[pier]/.run -g`
```
joemfb, line 397: This runs a gc pass at the end of every virtualized computation. Intended as a developer/qa tool for verifying refcounts. Currently only works if the binary was built with `MEMORY_DEBUG` (which changes allocation structures and produces incompatible snapshots).

content/using/running/vere.md
```diff
@@ -0,0 +1,633 @@
...
+
+- Undocked: `urbit --https-port 8443 [pier]`
+- Docked: `[pier]/.run --https-port 8443`
+
+### `-I, --inject FILE`
+
+Inject the event stored in `FILE` jamfile.
+
+- Undocked: `urbit -I /path/to/file.jam [pier]`
+- Docked: `[pier]/.run -I /path/to/file.jam`
```
joemfb, line 437: This event has "full privileges" and can do *anything* to your ship. `-I` is for customization in hosting environments, and emergency recovery for sinking ships -- only run from trusted sources.

content/using/running/vere.md
```diff
@@ -0,0 +1,633 @@
...
+
+### `-j, --json-trace`
+
+Create JSON trace files in `[pier]/.urb/put/trace`.
+
+- Undocked: `urbit -j [pier]`
+- Docked: `[pier]/.run urbit -j`
```
joemfb, line 447: This runs a tracing profiler, and produces JSON output compatible with `chrome://tracing`.

content/using/running/vere.md
```diff
@@ -0,0 +1,633 @@
...
+
+### `-P, --profile`
+
+Enable profiling. Profiles will be saved to
+`[pier]/.urb/put/profile/[now].txt`. This is useful for debugging and
+development. Profiling hit counters in Hoon code specified with the `~$` rune
+will be recorded here.
+
+- Undocked `urbit -P [pier]`
+- Docked `[pier]/.run -P`
```
joemfb, line 515: This runs a sampling profiler and produces custom output. Runtime cost metrics are only calculated if the binary was built with `CPU_DEBUG`.

content/using/running/vere.md
joemfb: (Answering the question here, not suggesting copy)

This argument disables hashboard, an alternate jet dashboard that validates jet registrations by the cryptographic hash of the battery. This option is always on (hashboard is always off), so it has no effect.

Hashboard itself works fine (although it can be slow in certain circumstances), but it implies a battery hash distribution channel that has never existed. It will most likely be replaced by other, newer jet dashboard designs.

joemfb (changes requested):
Thanks, this is great!

Going forward it might be best to keep old versions of this document around, indexing into the runtime reference by version number. And in future versions, I think we should more strongly privilege the docking interface, rather than present each option all the time.

I'd also prefer not to document the no-op arguments here. There are lots of them, and i don't think anything is gained by including them.

PR #1550, opened 2022-06-08
add unit test guide

tinnus-napbus:
brief guide to unit tests with the `-test` thread


drbeefsupreme:
Would be good to have an example of running a particular test or set of tests in a file rather than all of them. Off the top of my head I can't remember the syntax...

sigilante:
There's a long article I wrote for dcSpark a few months ago that may be worth linking, but we need to figure out our third-party linking/reference policy.

https://medium.com/dcspark/writing-robust-hoon-a-guide-to-urbit-unit-testing-82b2631fe20a

PR #1553, opened 2022-06-10
Update 2022-06-16-developer-call-urbits-new-terminal-stack-tlon.md

thomas-kroes:


Fang- (approved):
As discussed out of band. Looks good to me!

PR #1556, opened 2022-06-12
Seemingly inaccurate info for %kick gift

leowini:
%kick takes a list of paths, not a (unit path).


matildepark:
@leowini Can you reapply this to urbit/developers.urbit.org?

PR #1559, opened 2022-06-13
schema: strip special characters from section ids

matildepark:
Matching our previous build engine's behaviour.

`https://urbit.org/docs/hoon/reference/zuse/2d_6#++ot:dejs:format` -> `https://urbit.org/docs/hoon/reference/zuse/2d_6#otdejsformat`

### DIFF
--- a/components/schema/Heading.markdoc.js
+++ b/components/schema/Heading.markdoc.js
@@ -11,7 +11,7 @@ function generateID(children, attributes) {
   return bottomChildren
     .filter((child) => typeof child === "string")
     .join(" ")
-    .replace(/[=?!><:;,+#^|$&~"*@\._%/]/g, "")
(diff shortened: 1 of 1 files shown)


components/schema/Heading.markdoc.js
```diff
@@ -11,7 +11,7 @@ function generateID(children, attributes) {
   return bottomChildren
     .filter((child) => typeof child === "string")
     .join(" ")
-    .replace(/[?]/g, "")
+    .replace(/[=?!><:;,+#^|$&~"*@\._%/]/g, "")
```
tinnus-napbus:
```suggestion
    .replace(/[=?!><:;,+#^|$&~"*@\.%/]/g, "")
```

tinnus-napbus (changes requested):
I forgot it also retained underscores. Looks good otherwise

Issue #1543, opened 2022-06-02
Sidebar lists do not properly signal that there is more content


ynx0:
Good examples of this are https://urbit.org/faq and https://urbit.org/docs/hoon/reference/stdlib/2b

On page load, there is a scrollbar but it disappears fairly quickly. I will note that I'm using the Firefox browser on NixOS.

**Page-load view**
![image](https://user-images.githubusercontent.com/9275716/171696258-af9035a4-9fa3-4258-ab9a-43fddb3f33b8.png)

**After Scrolling**
![image](https://user-images.githubusercontent.com/9275716/171696643-a223ee3a-2740-4984-99ee-28470320df39.png)

The only way you can tell is that 3 pixels are missing from the bottom of the p, which is pretty easy to miss at that font size.

PR #1568, opened 2022-06-14
TableOfContents: sidebar niceties

matildepark:
- Category titles now highlight as you browse, not just child headers
- Sidebar now hides itself if there's only one heading on a page
- Sidebar fades to white when sidebar content is long (like on the FAQ, `/docs/hoon/reference/stdlib/2b`, etc.) (#1543)


gordoncc (approved):
Nice design polish!

PR #1570, opened 2022-06-14
Give explicit example for S3 setup

zalberico:
* I've seen multiple people confused trying to setup S3.
* AWS no longer supports v2 and doesn't work.
* DigitalOcean works, but has different names for things.
* Giving an exlicit example with images will make this easy for people.


matildepark:
Can we please stop committing images to the git repo? It's bad practice. You should upload to the media.urbit.org bucket.

zalberico:
Where is that bucket and how do I access it?

matildepark:
If you have access to the Tlon org it's [this bucket](https://console.cloud.google.com/storage/browser/media.urbit.org).

zalberico:
Is there a specific directory in there it's supposed to go?

matildepark:
Usually site folder, usually mimicking the urbit.org site folder structure. So /site/using etc

zalberico:
Looks like I don't have permissions to create a using folder in site/ (or os folder within using)

matildepark:
Made the folders, but if you can't make folders I'd be doubtful you could upload files.

zalberico:
Yeah, looks like I have read only access

content/using/os/s3.md
```diff
@@ -12,15 +12,50 @@ To enable S3 support on your ship you'll need to first set up a bucket, and then
...
+- Select the datacenter region closest to you
+- Restrict File Listing (leave this as the default)
+- Choose a unique name (this is your bucket name)
+- Select a project (or leave the default if you don't have multiple)
+
+Once you've created your space you'll want to navigate to its settings in DigitalOcean.
+
+Then you'll want to set the CORS Configurations as specified above, this should look like the following image (your endpoint will vary depending upon region).
+![digitaloceans3](../../../public/images/digitaloceans3.png)
+![](../../../public/images/s3cors.png)
```
tinnus-napbus: need to upload to media.urbit.org s3

tinnus-napbus (changes requested):
LGTM, just need to remove images from here & upload to the s3

content/using/os/s3.md
zalberico: I'm trying to get write access to this - do you have it? Are you able to give it to me?

tinnus-napbus: I have uploaded them. I have write access but not admin to give access:

`https://media.urbit.org/docs/user/digital-ocean-cors-settings.png`

`https://media.urbit.org/docs/user/digital-ocean-settings.png`

tinnus-napbus:
oh I should have probably put those in that folder you created, didn't read scrollback

tinnus-napbus:
have reuploaded to the correct folder:

`https://media.urbit.org/site/using/os/digital-ocean-cors-settings.png`

`https://media.urbit.org/site/using/os/digital-ocean-settings.png`

zalberico:
Sweet, thanks @tinnus-napbus - I made the requested change

tinnus-napbus (approved):

matildepark:
FYI ideally you should force push your branch so the binaries aren't in the tree. Since it doesn't store the diff but the full binary for the file it basically makes the .git download about as big as every single version of these image files, on or off the repo, when you add or remove them, and it scales very poorly. But it's too late now, so for future reference ...

PR #1571, opened 2022-06-15
Assign symmetric routing and notification protocol apprenticeships

mikeosborne:


thelifeandtimes (dismissed):

mikeosborne:
@thelifeandtimes what needs to be edited?

thelifeandtimes (commented):
sorry for my cryptic note, what I think requires an edit is that the `@p`s don't highlight or link out when we have two in the field. might be something @matildepark can easily fix? if not, we can merge without and just work on that in the future as we rework some of the metadata stuff.

thelifeandtimes (approved):
Approving and merging so as to not slow things down, but will add as an issue that the ideal is each of the assignees get individually parsed and hyperlinked.

PR #1572, opened 2022-06-15
components: cleanup omnibus

matildepark:
- Migrate some components into applicable folders to mimic the site directory
- Event patp mentions now link to ID pages too
- Event cards truncate the description after 1 line to keep from overflowing
- Allow background images to be darkened in the metadata

before / after
Screen Shot 2022-06-15 at 2 01 16 PM

Screen Shot 2022-06-15 at 2 02 53 PM


matildepark:
Also cleaning up a dangling `++` on the description of one of these pages that looked messy.

PR #1573, opened 2022-06-15
search: use Levenshtein distance algorithm

### LINKED ISSUE #1528: search: implement smarter algorithm
In particular we should sort incoming queries by [Levenshtein distance](https://en.wikipedia.org/wiki/Levenshtein_distance).

matildepark:
Fixes #1528

@ynx0, wanna try it out with me? Your examples in #1523 seem better now. We also return matches for `parent` folders too now so `Distribution` works and actually provides results.


ynx0:
Just took a look and "distribution" now working is quite nice.

Some others I finally got around to testing:
- cat/snag used to work and still work
- dif now works
- `tap` almost works, but `stap` is prioritized over `tap:fe`. IMO an exact prefix match should be prioritized here, but could be tricky to do write and might mess up other results if we're not careful
- alas/limb/time now work


"Software Distribution" doesn't work for just distribution; maybe doing a second query where instead of `and`ing the terms they are also `or`ed and these matches are put below all `and`ed results could work here.

A 1-deletion for something such as `dstribution` for distribution, or `tme` for `time`, yields no results; it would definitely be nice to help the user out here.

A nice tangent is the [Damerau-Levenstein distance](https://github.com/mrshu/node-weighted-damerau-levenshtein) but this is just something to look at rather than a recommendation.


With that said, I don't mean to hold back merging in the PR and I think it's good as it is.

matildepark:
> tap almost works, but stap is prioritized over tap:fe. IMO an exact prefix match should be prioritized here, but could be tricky to do write and might mess up other results if we're not careful

I was thinking about this exact case and only just realised it's because `stap` is 1 character off and `tap:fe` is two. I think we could prioritise `title.startsWith()` or whatever in the case of glossary entries just to cover that case better?

matildepark:
Creates too many weird cases ... indeed, will merge as is.

PR #1575, opened 2022-06-16
build: use foundation-design-system for all styling, markdown

matildepark:
The Foundation is about to have a lot of sites with the same styling on separate repos:

- [urbit.org](https://urbit.org)
- [operators.urbit.org](https://operators.urbit.org)
- [developers.urbit.org](https://developers.urbit.org)

And it already has other sites with disparate styling (eg. [network.urbit.org](network.urbit.org)).

To prevent excessive duplication of common components and enforce similarity in architecture and design, I've pulled out all our Markdown parsing, rendering, components; as well as all our CSS styles and Tailwind configuration into the [foundation-design-system](https://github.com/urbit/foundation-design-system) which allows for easy import in other sites. Instructions for usage and build are in that repo. For Markdown users, I've also prepared [this guide](https://github.com/urbit/foundation-design-system/blob/main/MARKDOWN.md) for pointers on using our components.

In this PR we just delete all the spare files and import from the upstream repository.


gordoncc:
Very cool.  Thanks for putting together the markdown documentation!

matildepark:
*shrugs*

PR #1582, opened 2022-06-17
Update meetup-bom.md to completed

mikeosborne:


thelifeandtimes (changes requested):
Let's change the reward count so that the metadata maps to actual final/total payouts?

thelifeandtimes (approved):

PR #1585, opened 2022-06-21
Hosting guide changes

tinnus-napbus:
it turns out DO lets you paste in a script to configure an instance when creating, so I've simplified a lot of the server config by having the user paste that in. It creates a new user, configures firewall, configures ssh keys, installs caddy and tmux, and installs the urbit binary.

After that it has the user sign up for a free subdomain, and then ssh in and configure caddy so it'll work and they'll have HTTPS.

I haven't touched the rest of the stuff about backing up and copying the pier and booting it. That stuff will need to be restructured around this flow


tinnus-napbus:
whoops PR'd to wrong repo

PR #1591, opened 2022-06-23
Create nested-goals.md

thelifeandtimes:
@mikeosborne can you double check me on meta tags?

@tacryt-socryp can you confirm your status as champion on this?

@niblyx-malnus can you confirm that this all looks like an accurate translation from your posts in the proposal group?


mikeosborne (approved):

tacryt-socryp:
Confirmed as champion.

niblyx-malnus:
Yeah this looks good!

PR #1598, opened 2022-06-28
Update index.js

gordoncc:
removed "early bird tickets" from callout since they are gone


matildepark (approved):

matildepark:
Don't Panic, know where your towel is, etc

PR #1601, opened 2022-06-29
Cancel Feed Aggregator project

mikeosborne:
Project replaced by trill-ufa.md


thelifeandtimes (approved):
Approved contingent on answering the question: how does this appear on urbit.org if cancelled? does the page disappear?

PR #1611, opened 2022-07-01
fix eyre-scry apply button

thelifeandtimes:
@tacryt-socryp can you confirm this works for you in the test deployment?


tacryt-socryp:
There is an Apply Now button and it goes to an Airtable. Looks good as far as I can tell! Thanks!

tacryt-socryp (approved):

PR #1623, opened 2022-07-07
remove chat-cli guide from user manual

### LINKED ISSUE #1056: remove chat-cli commands from 'using your ship'
The entire "join flow" as depicted doesn't represent the current chat-cli experience; you would use chat-cli to view chats you're in, because we don't join chats directly this way anymore, we join groups.

tinnus-napbus:
chat-cli is more-or-less deprecated and the guide in the user manual will just confuse people

resolves #1056


sigilante (approved):
Agree.

PR #1625, opened 2022-07-07
Update funes-notes.md

thelifeandtimes:
@matildepark  made some updates to this; if you can review and merge that would be rad.


matildepark (approved):
I feel like this is better facilitated by having grants have a "deliverable" section that links to the work than to link back this way all the time, but sure.

thelifeandtimes:
@matildepark agreed. @gordoncc and I discussed something along these lines about a month ago. definitely would be good to implement.

PR #1650, opened 2022-07-20
fixing metadata and copy link design glitches

gordoncc:
groups, ids, and applications metadata looks better now


matildepark (approved):
sure

PR #1655, opened 2022-07-21
move /using to operators.urbit.org

matildepark:
Part 1 of a 2 part deploy — the other side is in the [operators.urbit.org](https://github.com/urbit/operators.urbit.org) repo.

Removes all /using content, links to /using on urbit.org, and redirects all pages to the right page on operators.urbit.org.


gordoncc (approved):
hopefully not too many links break with this.

matildepark:
> hopefully not too many links break with this.

I did do a pass and the redirects should be okay. It’s just old redirects that might be the issue.

PR #1651, opened 2022-07-20
add ecosystem section

matildepark:
Fixes #1642 

Adds layouts for the ecosystem section featuring organisations, marketplaces, applications, podcasts, and a monthly spotlight. 

Showcases the spotlight content in custom cards on the front page.

Organization, podcast, marketplace pages all bidirectionally link when mentioned in a spotlight.

Ecosystem spotlights, because they need to render differently in different places, are far more table-like than post-like. Please see the following as a reference:

```
+++
// title just needs to be month
title = "June 2022"

// date required for sorting
date = "2022-06-01"

// featured sections aren't enforced in threes, but might produce weird layout
[featured-1]
title = "Urbit.live"

// description is used on index page for the card
description = "One line about the thing"

type = "Marketplace" // should be Marketplace, Application, Organization, Podcast

// images should be card-sized and will be sized down and filled as a cover in the icon'd, main ecosystem page
image = "https://s3.us-east-1.amazonaws.com/haddefsigwen1/reciprocal/2022.5.20..19.52.50-B137D4AA-DA9E-4B07-98CE-C001C576C7CA_1_201_a.jpeg"

URL = "https://urbit.live"

// content is parsed as markdown, but needs to be inputted here instead as a long-form string, using triple quotes
content = """
This layer 1 marketplace is all the rage.

Let this be a lesson to you."""

[featured-2]
title = "Reciprocal Ltd."
description = "One line about the thing"
type = "Organization"
image = "https://media.urbit.org/site/ecosystem/podcasts/full_1639502000-artwork.jpg"
URL = "https://reciprocal.ltd"
content = """
A design consultancy and friend network.

Let this be a lesson to you."""

[featured-3]
title = "Reciprocal Ltd."
description = "One line about the thing"
type = "Organization"
image = "https://media.urbit.org/site/ecosystem/podcasts/full_1639502000-artwork.jpg"
URL = "https://reciprocal.ltd"
content = """
A design consultancy and friend network.

Let this be a lesson to you."""
+++
```

Please take this and fill it with content (removing obvious sample content when necessary).

PR #1669, opened 2022-07-25
Update Header.js

gordoncc:
Part 1 of 2 - remove faq from header on urbit.org, add relevant sections to operators.urbit.org


matildepark:
FAQ's already removed in #1651 though?

PR #1680, opened 2022-07-28
Add chess.md

ashelkovnykov:
Had to manually add file - the auto-generated link from https://urbit.org/applications/submit was broken in multiple ways:
1. Had to touch each field before the submit button would be activated
2. Once active, the submit button would take me to a generalized failure landing page instead of making a new PR


matildepark:
For whatever reason if you have a pre-existing urbit.org fork it hits an error. I don’t even know where to report that to GitHub so there we are.

matildepark (approved):

PR #1684, opened 2022-07-28
Change title

sigilante:
I'm director of developer experience, not technical director, for whatever titles may be worth.


gordoncc (approved):
Thanks for clarifying

PR #1690, opened 2022-08-02
Create the-shape-of-dao-governance-to-come.md

gordoncc:


matildepark (approved):
Sure? My main thought here is that this is the third Combine cross-post on the blog in a row

jalehman:
Agree with @matildepark here. We need other content coming onto the blog so as to not completely hijack this.

gordoncc:
Ha, I've got another guest blog post that just needs some editing on Urbit and IoT but I thought it would be cool to have a neat consecutive series on DAOs.  Happy to break it up a little though.

PR #1693, opened 2022-08-04
theme: add dark mode

matildepark:
- Bumps foundation-design-system to 0.2.0, which adds a requirement for `tailwindcss-theme-swapper` and dictates a dark theme. The dark theme inverts the order of all colours (green-100 becomes green-600) and uses antiWall palette instead of our wall palette for the same weight.
- Finally figured out how to require modules from a parent explicitly, so we don't have to declare tailwindcss plugins and can defer to FDS entirely
- Was more explicit with styles when necessary — sometimes no background meant `white` on both themes
- Changed all svgs to use css fill classes instead of explicit fill colours so they toggle.

Dark theme currently uses the system preference. We can add the toggle as a separate pass, I just think it should be done sanely (imagine jumping from site to site and it toggles to light mode each time...).

Screen Shot 2022-08-03 at 5 43 05 PM
Screen Shot 2022-08-03 at 5 43 16 PM


matildepark:
The build cache wasn't stale even with a breaking change to imports ... a build after my own heart ... 

Presumably #1689 worked ...

gordoncc:
Great start.  Love the inversion on the SVGs.

Made some slight modifications preserving the original color palate, see homepage mockup here: https://www.figma.com/file/rlQkryJk1WjzTnMNZzsAkl/Urbit.org?node-id=188%3A731

See the background, the buttons, links, and assembly card.

Basically let's not invert the buttons and lighten up the background a bit.

gordoncc:
Edit – let's just start by changing the background from pure black to a shade like #151618

May want to do some more tweaks later but for now I think that will be good enough to get us rolling.

gordoncc (approved):

PR #1714, opened 2022-08-12
pages: add decentralized-discord landing page

matildepark:


gordoncc:
Looks like the video playback isn't working on mobile unless there's some trick I'm not aware of

matildepark:
[Turns out](https://developer.apple.com/documentation/webkit/delivering_video_content_for_safari):

> Video elements that include `<video autoplay>` play automatically when the video loads in Safari on macOS and iOS, only if those elements also include the playsinline attribute.

gordoncc:
Nice, playback working for me now.  Is there a way to make it loop or would we need gif technology for that

gordoncc:
Ok, I'd just like to get a custom opengraph card image in and we're good to deploy:

https://storage.googleapis.com/media.urbit.org/site/landing/discord-opengraph.png

matildepark:
image

gordoncc (approved):
Thanks!

PR #1723, opened 2022-08-16
Create Faux app listing

thelifeandtimes:
@midsum-salrux can chime in with additional edits if desired.


midsum-salrux:
https://github.com/midsum-salrux/faux

could you set this as the website? otherwise looks good.

gordoncc (approved):

PR #1730, opened 2022-08-16
landing: design additions after feedback

matildepark:
Adds the blurbs, screenshots, and revisions as per Figma. Tested in both mobile and on dark mode.


gordoncc (approved):

gordoncc:
Looks good!  One nice to have feature would be be clicking on the screenshots opens them in larger size modal window, but we can go ahead and push this.

PR #1740, opened 2022-08-17
adding port beta notice to desktop app instructions

gordoncc:
getting lots of feedback from newbies that they can't connect to groups or install apps with port, need to manage expectations.

port notice

### DIFF
--- a/content/getting-started/desktop.md
+++ b/content/getting-started/desktop.md
@@ -4,6 +4,9 @@ description = "Run Urbit locally as a desktop application"
 weight = 1
 +++
 
+<div class="bg-yellow-200 rounded-xl display !p-7 flex flex-col space-y-4"><h4 level="4" id="notice">Notice</h4><p className="text-sm">Port is early beta software, and inconsistent functionality has been reported in certain operating environments.  A more reliable way to quickly check out the network is in the works.</p><p>For the best immediate Urbit experience we recommend <a href="/getting-started/server">setting up a VPS</a> or <a href="/getting-started/hosted">using a hosting service</a>.</p></div>
+
+
 ### 1. Get Port
 
 Port is a desktop GUI application that allows you to spin up, access, and manage your Urbit ships without any knowledge of the command line or Urbit binary.
--- a/content/getting-started/desktop.md
+++ b/content/getting-started/desktop.md
@@ -4,6 +4,8 @@ description = "Run Urbit locally as a desktop application"
 weight = 1
 +++
 
+
+
 ### 1. Get Port
 
 Port is a desktop GUI application that allows you to spin up, access, and manage your Urbit ships without any knowledge of the command line or Urbit binary.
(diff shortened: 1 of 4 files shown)


matildepark:
I would be specific: it's really just Windows.

matildepark (commented):
Don't we have `{% callout %}` for this?

gordoncc:
Yea, but I wanted it to add in that yellow background.  Can I pass a class to the callout?

matildepark:
The colour is hardcoded on the component right now, but we could always change that to just be a fallback. You can also use `{% div %}` here, but it seems moot. I'm just averse to introducing more inline html.

matildepark:
My other qualm would be "does this introduce a contradiction to other copy in this section" — don't we recommend Port as a go-to? This is a bit of a mixed signal and might indicate a need to do something bigger here with the strategy.

gordoncc:
Would be nice to be able to pass classname so we could have different types of callouts - alerts, heads up, etc - and have the gray as a fallback.

gordoncc:
Yeah I'm with you on that qualm -- Port should be replaced by a pre-mined comet that you access in browser.  Wasn't @pkova working on something like this?

matildepark:
Also "I can't install apps or join groups" smells like networking and/or networking keys issues to me. The thing where people get mysterious error message when they open ship in Windows is Port-specific but you know.

matildepark:
Okay, try now. You can do `{% callout className="bg-yellow-200" %} Content content {% /callout %}`.

matildepark:
It's working — but Tailwind is not adding the class to the CSS and detecting that it's there even though it does elsewhere. Let me see what I have to adjust.

gordoncc:
Screen Shot 2022-08-29 at 8 25 35 PM

Relocated this warning but for some reason the background color doesn't show up here, but it does locally.  Is this an FDS issue or something like that?

gordoncc:
Oh just saw your comment, you read my mind

matildepark:
So we usually pass Markdown content into Markdoc to transform it first so Tailwind can see what a class is; but in this case because we aren't using FDS's custom components, the sections are skipped in the transform. Using FDS's renderer directly makes Tailwind upset in a way I don't currently understand. Removing the transform provides normal behaviour; I looked at other Markdoc injections of classes in the urbit.org codebase and they appear fine (when I would expect they wouldn't — like sure it knows what `className=""` is but why does it know `.font-bold` etc. in the inline declarations in Overview?). A little confused.

gordoncc:
Is it ok to merge or should we do some more testing to make sure nothing is broken?

matildepark:
I skimmed through all mentions of Markdoc components and it looked clean. We don't have that much content on urbit.org anymore.

gordoncc:
Well, there's no rush, there's already a warning on the Windows section anyway so this is just styling.  Curious about the tailwind / fds issues.

matildepark:
Kept the steps as is as before, but using FDS's parser instead of Markdoc's. This should produce least amount of potential failures, presumably, but again I looked at all our custom Markdoc calls with class declarations and they seem to be in place.

gordoncc:
Let's go for it

matildepark (approved):

PR #1744, opened 2022-08-17
pages: update opengraph cards

### LINKED ISSUE #1708: Update Opengraph Card Formatting
Updated Opengraph cards, should be applied across the sites.

https://www.figma.com/file/rlQkryJk1WjzTnMNZzsAkl/Urbit.org?node-id=198%3A1405

Opengraph Cards

matildepark:
- We now deploy a server-side generator of sigil .pngs at `https://urbit.org/api/sigil` that takes `patp` (with a leading `~`) and `color` (without a leading `#`) and outputs a 1024x1024 png.
- All pages except events and blog posts use `summary` and not `summary_large_image`.
- The homepage now actually has OG data. It did not before. This is crazy to me.
- All OG cards no longer have `- urbit.org` suffixes, instead declaring what they are.
- All pages in organisations and marketplaces now have descriptions. Podcasts default to describing themselves in cards in `Podcast - Date` format. This is all necessary because we use Markdown in the descriptions of all this content and we need a text-only snippet for cards.

ID pages do in fact look correct on Twitter, but they won't right now because Twitter images need to be absolute paths, so we point to the non-existent Sigil api on prod. Here's what it looks like:

Screen Shot 2022-08-17 at 12 39 35 PM

Fixes #1708

### DIFF
--- a/components/BasicPage.js
+++ b/components/BasicPage.js
@@ -22,7 +22,7 @@ export default function BasicPage({ post, markdown, search, index = false }) {
   return (
     <Container>
       <Head>
-        <title>Urbit • {post.title}</title>
+        <title>{post.title}</title>
         {Meta(post)}
       </Head>
       <IntraNav ourSite="https://urbit.org" search={search} />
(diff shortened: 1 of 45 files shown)


jalehman:
Huge

gordoncc (approved):
sweet

PR #1746, opened 2022-08-17
Create 2022-09-01-developer-future-of-khan.md

thomas-kroes:


jalehman (changes requested):
Filename should be changed to remove the word holium and the % sign.

thomas-kroes:
Done, that's what happens if you rush it.

jalehman:
The filename needs to be changed, not the PR name. The filename becomes the URL.

thomas-kroes:
argh, done.

matildepark:
We should probably host this on media.urbit.org and not imgur — it's a little finnicky with hotlinks. @gordoncc ?

thomas-kroes:
@matildepark I agree. @gordoncc told me to do that. I still don't have access to that storage, so I opt for the easiest possible route.

jalehman:
Now you do

thomas-kroes:
Thanks Josh, should be ready to merge now. @gordoncc we might want to look at the image again for the next one as the contrast with the grey text is not amazing.

jalehman (approved):

PR #1751, opened 2022-08-21
added sortug organization

polwex:


matildepark:
Both the URL and the image don't work.

thelifeandtimes:
Feels weird to have so many ships listed?

matildepark:
Why does it matter how many ships are listed? Those are used to check for applications and groups the org owns.

thelifeandtimes:
ah, I see.

polwex:
My apologies, the logo and website are fixed, should be working now.

thelifeandtimes (approved):
lgtm

PR #1739, opened 2022-08-17
grants: allow multiple grantees

matildepark:
Re-implementing the work in the `multiple-grantees` branch that wasn't used before. 

Fixes #1578

PR #1753, opened 2022-08-23
enable arrays for mentors and champions

thelifeandtimes:


thelifeandtimes:
@matildepark build issues here.

matildepark:
As of #1739 assignees look like arrays eg. `["person1", "person2"]`.

thelifeandtimes:
> As of #1739 assignees look like arrays eg. `["person1", "person2"]`.

and per my change I take it this doesn't work for champions?

matildepark:
First, you did it like `["person1, person2"]` but it's two quoted names, comma-separated. Second, no, champions aren't currently set up that way.

thelifeandtimes:
would that be simple to implement?

thelifeandtimes:
@gordoncc this should be ready for review and merging

thelifeandtimes:
@thomasurbit met with the gather grantees and they are going to make further changes, so pulled the grant content out of this PR. merging should just add the mentor/champions as array feature.

gordoncc (approved):

PR #1761, opened 2022-08-25
Create persistent-memory.md

thelifeandtimes:
@eamsden here's a draft of this; please make commits directly to this branch for milestones and let's talk about funding.


eamsden (commented):
It looks good to me.

content/grants/persistent-memory.md
```diff
@@ -0,0 +1,158 @@
...
+## Milestone 1: Basic copy-on-write arena
+
+Implement a copy-on-write arena by using `write()` to copy pages to new regions of the backing file.
+This should include at least single-arena snapshot support.
+
+Compensation: 1 star
+
+## Milestone 2: Kernel copy-on-write (`MAP_PRIVATE`) + lazy write arena
+
+Implement a copy-on-write arena by mapping pages `MAP_PRIVATE` and lazily writing dirty pages to the backing file.
```
joemfb, line 142: It might be worth noting that this is will shortly be vere's approach.

content/grants/persistent-memory.md
```diff
@@ -0,0 +1,158 @@
...
+
+Compensation: 1 star
+
+## Milestone 3: Memory allocation and separate event arena
+
+Implement a memory allocator on top of this arena. This includes the ability to store metadata such as roots which will be committed with the snapshot, and reference counting or tracing to free nouns which are not referenced from the root.
+
+Implement a separate (may be interleaved at the page level in memory/on disk) arena for event logs, which is bump-allocated and not copy-on-write but rather does not reallocate in committed pages.
+
+Compensation: 2 stars
```
joemfb, line 152: I'm more inclined to having a milestone for making the "basic copy-on-write arena" less basic (adding a b+tree, maybe optimizing copy destinations for adjacency so as to minimize page table entries, or anything else that comes up). I'm not sure how much the details of the actual allocator/arena will be tied to and require coordination with the larger new-mars project.

joemfb (approved):
If this design is a success in practice, I'd be very interested in back-porting it to vere. I've noted a couple relevant points. But this looks good regardless.

content/grants/persistent-memory.md
eamsden: There is a difference in that the specified approach for committing here is to find free blocks and `write()` to them, then update the index. Vere `write()`s to a patch file which is later unified with the main snapshot memory image, correct?

content/grants/persistent-memory.md
eamsden: I think something like a B tree or B+ tree for the index is a necessary part of the initial implementation actually. But it's also not hard and can be implemented in that few-weeks timeframe.

joemfb: Ok, sounds good. I wasn't sure just how minimal we were talking.

content/grants/persistent-memory.md
joemfb: Interesting, I hadn't thought about the space between the main proposal and vere's approach. Don't mind me ...

PR #1770, opened 2022-08-26
added trill and ufa apps

polwex:


thelifeandtimes:
@matildepark what is the syntax / how do orgs get linked from applications? is it based on the development ships listed or something else?

matildepark:
> @matildepark what is the syntax / how do orgs get linked from applications? is it based on the development ships listed or something else?

If the developer is a ship associated with an organisation, it shows a link to the organisation instead.

(However, here it's hard to test because the OP's fork of urbit.org precedes the merge for Sortug Development, so that organisation isn't on the deployment preview of the site. So you'll have to just list it as ~sortug etc. and see it on production afterward.)

gordoncc (approved):

PR #1775, opened 2022-08-30
Add September 2022 Ecosystem Spotlight

gordoncc:
Added new spotlight which displays properly on /ecosystem, but the old content is still showing up on the homepage index.  @matildepark can you update so that the latest ecosystem spotlight displays on the homepage cards?

### DIFF
--- a/content/ecosystem/spotlight/2022-september.md
+++ b/content/ecosystem/spotlight/2022-september.md
@@ -32,6 +32,8 @@ Peat is Urbit’s first backup utility.  Prior to Peat, there was no easy way to
 
 title = "Josh Lehman: Urbit – The Peer-to-Peer Network Re-Decentralizing the Internet"
 description = "Urbit Foundation Executive Director Josh Lehman and Epicenter host Brian Crain have a wide ranging discussion on Urbit's history and future."
+image="https://storage.googleapis.com/media.urbit.org/site/ecosystem/spotlight/epicenterjoshthumb.jpg"
+
 content = """
 
 In this podcast Urbit Foundation Executive Director Josh Lehman provides a broad overview of that state of Urbit as of 2022, and what people can expecture in the near future and beyond.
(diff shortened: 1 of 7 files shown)


content/ecosystem/spotlight/2022-september.md
```diff
@@ -0,0 +1,42 @@
...
+
+You can currently submit and search for applications, groups, and content (notebooks and collections), as well as browse by tag.
+
+![Screenshot](https://storage.googleapis.com/media.urbit.org/site/ecosystem/applications/sphinx2.png)
+
+"""
+
+[featured-2]
+title = "peat"
+description = "Need a personal website, fast? Studio can set one up in under a minute."
```
matildepark: Needs a description

gordoncc:
Looks like we need to account for apps that only use background colors on these cards.
Screen Shot 2022-08-30 at 1 38 49 PM

matildepark:
We do (in Safari) — 
image

Chrome layout engine discrepancy, let's see what the difference is

gordoncc:
Hmm it doesn't look great w/o an image, I should probably just put a screenshot there.

gordoncc:
Ok, a few more issues – there's no background color on the sphinx image in the card, and when I add an image for a card it appears in the spotlight icon thumbnail.  Maybe that was deliberate but we should take that out, then we can use custom images for the index cards.

Screen Shot 2022-08-30 at 2 07 53 PM

matildepark:
Can you push?

gordoncc:
Pushed

gordoncc:
Looking much better, except adding a custom image replaces the post image now ... anyway we can get that original podcast/app image in there and keep the custom image on the index card?

Screen Shot 2022-08-30 at 4 10 16 PM

matildepark:
That's how it was intended to be written but I dropped one. Should be fine now.

gordoncc:
lgtm

matildepark (approved):

matildepark:
Merge or hold until Sep?

gordoncc:
It's called we release just a little ahead of schedule, like magazines

PR #1778, opened 2022-08-30
index: flatten assignee arrays for counting unique contributors

matildepark:
While console.logging this, found a bug — we weren't flattening arrays before deduplication so it was deduplicating the entire arrays against each other instead of their items. Our contributors were about 40 too high.


thelifeandtimes:
Thanks for catching this.

gordoncc (approved):

PR #1780, opened 2022-08-31
added assembly group

gordoncc:
/groups/~bitdeg/assembly-22


content/groups/~bitdeg/assembly-22.md
```diff
@@ -0,0 +1,10 @@
++++
+title = "Assembly 2022"
+shortcode = "~bitdeg/assembly22"
```
matildepark: This is different from the filename. Is this the right code? Could you amend them to match?

gordoncc: amended

matildepark (approved):

PR #1784, opened 2022-09-01
pages: add sigil generator

matildepark:
In which we speedrun writing our own https://sigil-generator.vercel.app. 

Screen Shot 2022-08-31 at 11 32 47 PM

@gordoncc What should we be doing with the copy on this page?

### DIFF
--- a/components/Sigil.js
+++ b/components/Sigil.js
@@ -12,11 +12,19 @@ export const foregroundFromBackground = (background) => {
   return whiteBrightness - brightness < 50 ? "black" : "white";
 };
 
-export const Sigil = ({ patp, size, color = "#24201E", icon }) => {
+export const Sigil = ({
+  patp,
+  size,
+  color = "#24201E",
+  foregroundColor = false,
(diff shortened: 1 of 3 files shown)


thelifeandtimes:
I'd suggest echoing some of the copy/messaging on this page: https://urbit.org/overview/urbit-id

edit: along with explanation that this is a sigil generator....

gordoncc:
Neat!

1.  This is clearly not the Urbit ID page, but its own Sigil Generator page.
2. This should be linked to from an Urbit ID page, which does not yet exist.  Someone should probably write and design that, I guess that someone is probably me.
3. /ids should probably redirect to an IDs section in the ecosystem.  Except it seems odd to call it "IDs" –   I guess we could call it "Ships" but that might even be more confusing.

matildepark:
I think it is cool anyway and I felt like doing it for fun.

thelifeandtimes:
Closing PR as stale / merged elsewhere.

PR #1786, opened 2022-09-02
Update faq re: grants program

thelifeandtimes:
Was looking at old gifts, found this page and tossed some quick updates into it. Not sure what points to it, aside from navigating directly to https://urbit.org/faq.

### DIFF
--- a/content/faq.md
+++ b/content/faq.md
@@ -136,15 +136,17 @@ Thus, while Urbit is probably more secure and private than most digital communic
 
 ### What are the different types of grants? {% #grants-1 %}
 
-There are three ways to receive a [grant of stars](https://grants.urbit.org/) for your contribution to the Urbit ecosystem.
+There are three ways to receive a [grant of stars](https://urbit.org/grants) for your contribution to the Urbit ecosystem.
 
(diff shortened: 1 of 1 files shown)


matildepark:
We kinda sorta retired it I think.

gordoncc:
Oh, I pulled out the technical guides into Operators.  I thought I removed the file, maybe I just removed the link.

Is there any grants program info in here that we need to display or can we just erase this?

matildepark:
This page is abandoned and outside the current IA, so we should just delete the page outright. Some of this content should be on the /grants page, maybe? But it seems kind of moot stuff ... maybe some mini-FAQ pages on the bounty pages.

thelifeandtimes:
post assembly i'll take another look and move some of this content off into /grants somewhere. assuming we have no plans to revive the FAQ more broadly? dunno if it makes sense to more a bunch of this into a footer item for operators.urbit.org, developers.urbit.org, and urbit.org/grants or something like that?

thelifeandtimes:
closing this out; grants page has been updated in a separate band.

PR #1790, opened 2022-09-08
updating homepage card for assembly with link to assembly 22 group

gordoncc:


pages/index.js
```diff
@@ -66,11 +66,19 @@ export default function Home({
...
               </p>
-              <Link href="https://assembly.urbit.org" passHref>
-                <a className="button-lg max-w-fit bg-green-400 text-white">
-                  Learn More
-                </a>
-              </Link>
+
+              <div class="flex flex-wrap">
+                <Link href="https://assembly.urbit.org" passHref>
+                  <a className="button-lg max-w-fit bg-wall-600 text-white mr-3 mb-8">
```
matildepark: Let's add `target="_blank"` to this. It sees urbit.org in the link so it isn't opening in a new tab, but it belongs outside the urbit.org family of sites IA.

pages/index.js
```diff
@@ -66,11 +66,19 @@ export default function Home({
...
+                    Learn More
+                  </a>
+                </Link>
+                <Link href="/groups/~bitdeg/assembly22" passHref>
+                  <a className="button-lg max-w-fit bg-transparent text-wall-600 border-2 border-wall-600">
```
matildepark: Maybe `hover:bg-wall-600 hover:text-transparent`? Looks weird as is imo but up to you.

matildepark (approved):
Fine otherwise

gordoncc:
good catches

PR #1794, opened 2022-09-12
Trial comets

pkova:
Adds a page for accessing trial comets `/trial`, and a static page where I can redirect people whose trials have ended `/trial-over`. Needs copy and possibly styling on both pages, backend is now good to go. Also, nothing links the trial landing page so we need to figure out where to put it. @gordoncc

### DIFF
--- a/pages/index.js
+++ b/pages/index.js
@@ -45,8 +45,9 @@ export default function Home({
           // Hero Statement
         }
         <Section className="pb-36">
-          <div>
+          <div className="space-y-8">
             <h1>A clean-slate OS and network for the 21st&nbsp;century.</h1>
+            <Link href="/trial" passHref><a className="bg-green-400 text-white button-lg max-w-fit">Explore the Network</a></Link>
           </div>
         </Section>
         {
--- a/pages/index.js
+++ b/pages/index.js
@@ -21,27 +21,6 @@ import fs from "fs";
 import path from "path";
 import Meta from "../components/Meta";
 
-const postReq = (path, params, method='post') => {
-  const form = document.createElement('form');
-  form.method = method;
-  form.action = path;
-
-  for (const key in params) {
-    if (params.hasOwnProperty(key)) {
-      const hiddenField = document.createElement('input');
-      hiddenField.type = 'hidden';
-      hiddenField.name = key;
-      hiddenField.value = params[key];
-
-      form.appendChild(hiddenField);
-    }
-  }
-
-  document.body.appendChild(form);
-  form.submit();
-}
-
-
 export default function Home({
   posts,
   ecosystem,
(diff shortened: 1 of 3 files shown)


matildepark:
There's something about this and #1784 that are currently not served in our websites — they're effectively app interfaces, not web pages, and I think they need their own sort of layout to them. I've tried to gesture at these in the past by making them "card-like", wrapping the whole thing but, I don't know, just calling attention to it.

The front-end looks pretty good! Definitely needs at least copy; should also be tied into getting-started section.

matildepark:
Er, as something me and Gordon can do I mean.

gordoncc:
Put together some designs here: https://www.figma.com/file/rlQkryJk1WjzTnMNZzsAkl/Urbit.org?node-id=304%3A994

@matildepark - think you can sort this?


Trial Concluded
Explore the Network - Not Available
Explore the Network
Index

matildepark:
Wow, synchronicity ...

gordoncc:
Looking good, just noticing that the button kinda breaks at smaller screen widths

Screen Shot 2022-09-13 at 10 36 42 AM

gordoncc:
One more feature request – can we turn the image into a carousel that rotates through images every 2s?  I feel like we need to do a little more showing on this page.
Explore the Network

gordoncc:
Here are some carousel images:

Carousel 1
Carousel 2
Carousel 3
Carousel 4

matildepark:
What if we just used the video from the discord page?

gordoncc:
Oh, even better idea.  Let's do that.

matildepark:
It's just like, to me, a carousel is like "ah, all these memories of Landscape .... they will be lost in time, like tears in rain ..."

gordoncc:
😅

gordoncc:
Flexbox is still having issues in Chrome

Screen Shot 2022-09-13 at 12 41 44 PM

gordoncc:
And the button /text should be left aligned rather than middle aligned (safari)

Screen Shot 2022-09-13 at 12 42 42 PM

gordoncc:
Ok, looking much better.  I made a new video that I think is a bit of a better demo, why don't we use this one: https://storage.googleapis.com/media.urbit.org/site/explorethenetwork.mp4

gordoncc (approved):

matildepark:
I guess we're good to deploy. Who has final stamp? @gordoncc @pkova ?

pkova:
We ran some numbers with @jalehman and decided to do a lot more comets per day but for just 24h. I'll make the necessary changes on the backend, fix the copy to say 24h instead of one week and then merge.

matildepark:
A day is brief; that’s a lot of comet turnaround. Can we confirm the subscription overhead it’ll add to bigger groups with an Ames / userspace person?

pkova:
I will do moons and breach them at the end of the trial period.

pkova:
And we are live. Fear the `rosmyn^fordet`.

PR #1810, opened 2022-09-24
Added urbit system diagram

zendevil:
So that newbies can get a higher-level picture of how Urbit works.

### DIFF
--- a/README.md
+++ b/README.md
@@ -56,3 +56,8 @@ We deploy using Vercel. All pull requests will show a deployment preview via the
 ## Upgrades
 
 The `/scripts` contains scripts used to programmatically upgrade site content data.
+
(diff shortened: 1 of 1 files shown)


zendevil:
Could somebody please review the PR? It's for helping understand how the Urbit network system works.

matildepark:
Here? On urbit.org's README? Seems like a weird place.

zendevil:
> Here? On urbit.org's README? Seems like a weird place.

Probably should go on the Overview page. Is the diagram correct? Could be added possibly in the Architecture section: https://urbit.org/overview/urbit-os. How does one upload images to media.urbit.org/site endpoint? Created an issue: https://github.com/urbit/urbit.org/issues/1821

PR #1825, opened 2022-10-03
adding assembly hackers group and quorum to apps

gordoncc:
Adding this content so we can link to it in the upcoming newsletter.


thelifeandtimes (approved):
I added some content, lmk if you agree it should be included.

gordoncc:
lgtm

PR #1584, opened 2022-06-21
site: add developers.urbit.org styling and components

matildepark:
Preparatory work for the new site's changes to the overall design of our websites. Might want to hold on a merge until developers.urbit.org ships, but at least there's a preview.

- `type-ui` now uses `font-medium` instead of `font-semibold`, slimming our navigations
- Adds an `IntraNav` component to the top of the website to link out to other ones, where Search now lives
- Pulls `Container`, `Section`, `Footer` and `SingleColumn` into [foundation-design-system](https://github.com/urbit/foundation-design-system) and generalises the Footer component to take an object of links and nav headers instead of hard-coded DOM elements

PR #1842, opened 2022-10-10
Update vercel.json

gordoncc:
redirect /docs/getting-started to /getting-started


matildepark (approved):
Went through and looked and yeah, every other docs directory redirects normally. Probably my fault during #1584, though I covered each directory one by one and wasn't thinking about docs folders a few iterations back in the site history.

PR #1851, opened 2022-10-20
grants: allow year subdirectories while preserving permalinks

matildepark:
- Moves all grant files into appropriate subdirectories by year.

<details>
<summary><code>initialBuildDatedGrants.js</code></summary>

```js
const fs = require('fs');
const matter = require('gray-matter');
const { join } = require('path');
const toml = require('@iarna/toml');

const options = {
    engines: {
        toml: toml.parse.bind(toml),
    },
    language: "toml",
    delimiters: "+++",
};

const dir = fs.readdirSync(join(process.cwd(), 'content/grants'), { 'withFileTypes': true });

return dir.filter((f) => f.isFile())
    .map((f) => {
        const file = fs.readFileSync(join(process.cwd(), 'content/grants', f.name))
        const { data } = matter(file, options);
        const year = new Date(data.date).getFullYear();
        const oldPath = join(process.cwd(), 'content/grants', f.name);
        const newPath = join(process.cwd(), 'content/grants', String(year), f.name)
        try {
            fs.readdirSync(join(process.cwd(), 'content/grants', String(year)));
            fs.renameSync(oldPath, newPath);
        } catch {
            fs.mkdirSync(join(process.cwd(), 'content/grants', String(year)))
            fs.renameSync(oldPath, newPath);
        }
    })
```
</details>

- For all grants pages, maps through years into a flat list to derive slugs.
- For all grants pages, searches through year subdirectories when trying to get post content, then returning the first directory with a file by that slug name.
- Adjust functions on grants.js, index.js, etc. interfaces that reference grants data to now map through year subdirectories and flatten.

In practice this just means that all links are the same, but grants are in year folders in the file directory.

In general, this sort of thing is an antipattern, so most of the logic is just written inline here (and it's fairly succinct), but if it becomes something we do a lot (that is, not matching filepaths to slugs), we'll probably have to refactor this sort of thing into functions.


matildepark:
There is one place this shows up elsewhere though, and it's in search. I don't really mind the appearance, but I could see how it could confuse.

image

This was in here before this PR, but we should probably sweep categories and make the titles match.

image

thelifeandtimes:
I actually don't totally mind it showing up in search. It will mean probably not doing it down to quarterly directories but this should give us an affordance of a bit more cleanliness. I'll go through and update the tags data in this branch to tidy stuff up.

thelifeandtimes (approved):

PR #1855, opened 2022-10-24
pages: fixing typo in power

solderneer:
I assume this was meant to say 2^8 not 28, I was mildly confused because 28 * 28 is not 2^16 :)


thelifeandtimes (approved):
thanks for catching. also caught another error.

PR #1887, opened 2022-11-08
metadatablock: always use 1/2 basis

### LINKED ISSUE #1885: Grants metadata cramped when multiple assignees present
image
Was looking at [this grant](https://urbit.org/grants/chess-bounty) last night and when the window narrows the metadata text tends to get incredibly cramped before moving to a newline or similar.

Expected behavior would be something along the lines of having each `@p` stay on its own line, and if width is overrun, then the field would move to a stacked column instead remaining on the same row.

matildepark:
Quick go at it. Fixes #1885

Any better @thelifeandtimes? This also affects app pages, group pages, ID pages etc but none of those have as many content in the metadata.


thelifeandtimes (approved):
Yup, this looks good.

gordoncc (approved):

PR #1890, opened 2022-11-10
Urbit Developer Week: s/Friday/Thursday for Core Dev AMA event

matthew-levan:
The meetup/youtube links say this event is on Thursday, but the prose doesn't reflect that.


thomas-kroes (approved):

thomas-kroes:
@matthew-levan thanks for flagging this, very helpful!

PR #1896, opened 2022-11-16
add bitcoin page

matildepark:
- Adds landing page at /bitcoin. Content for the page is located in a markdown file (bitcoin.md)
- Adds tags to Bitcoin posts and events
- Adds a query parameter for blog and event pages. When accessing to an index page with a tag query, it falls back to a new view filtering the posts down:

image

### DIFF
--- a/components/EventPreview.js
+++ b/components/EventPreview.js
@@ -41,11 +41,10 @@ export default function EventPreview({ event, className, big }) {
             )}
           >
             <div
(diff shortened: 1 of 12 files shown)


thelifeandtimes (approved):
LGTM.

gordoncc:
Looked good to me but we're missing the site footer, so let's add that back in.

matildepark:
Easily done, easily done …

PR #1898, opened 2022-11-16
pages/applications: remove related grants if none available

### LINKED ISSUE #1895: "Related Grants" heading appears on applications without related grants
Screen Shot 2022-11-14 at 3 49 57 PM

matildepark:
Fixes #1895

I was confused because I did write all the fallback logic to just hide the section; but it turns out the getGrantsByShortcode function I wrote falls back to an empty array, which is truthy, so we don't pass null, so there's *always* a `matchedGrants` value and thus it was showing. Since we know there's always the array, we check length of the array now.

### DIFF
--- a/pages/applications/[[...application]].js
+++ b/pages/applications/[[...application]].js
@@ -111,7 +111,7 @@ const ApplicationPage = ({ data, markdown, organisation, matchedGrants, search, 
             markdown={markdown}
           />
           <hr className="text-wall-200" />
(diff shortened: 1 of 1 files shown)


gordoncc:
Are there any applications with related grants?  I can't seem to find any

matildepark:
The only grant with `deliverable =` is Faux.

gordoncc (approved):
Thanks, looks solid to me

PR #1901, opened 2022-11-16
pages, components: pull Search into design system

matildepark:
Search was maybe our biggest, least scalable cross-site component. Because result order, result API calls, and result slugs differed per site, we had four copies of Search that *sort of* differed but not really?

Search was added to urbit/foundation-design-system 0.7.0 with an additional prop called `order` that is structured as an array and falls back to:

`["org", "dev", "ops", "roadmap"]`

The order is used to propagate what and in what order it gets posted to the search box.

Because we need to have "canonical APIs" that don't have an idea of what "our results" are, we add another API on each site for internal vs. external searches as a consequence.

This PR also:
- adds roadmap.urbit.org searches
- Pulls Sigil components into the design system too

Matching PRs for the other sites still pending.


gordoncc (approved):
Nice!

PR #1902, opened 2022-11-17
Set up of links between apps and grants

thelifeandtimes:


content/grants/2019/bitcoin-node-json-rpc-api-part-1.md
```diff
@@ -14,6 +14,8 @@ assignee = ["yosoyubik"]
 id = "649469235"
 completed = true
 link = ""
+link = "~pindet-timmut/urbitcoiners"
```
matildepark, line 16: It's failing because `link = ` is written twice on all these.

thelifeandtimes: ah, thanks.

thelifeandtimes:
@matildepark, i fixed those metadata issues but now is throwing a more complex set of errors.

thelifeandtimes:
@matildepark thanks for the fix. Can we exclude from the app pages any grants where `canceled = true`? those are excluded on the grants main page and I think they would make sense to exclude on app pages as well.

thelifeandtimes:
Also, would be good to sort the grants on the app pages with most recent at the top?

matildepark:
I can't see an example of any pages with these cases, but that should do it.

thelifeandtimes:
the bitcoin app is the one where I noticed both these things. This looks to have fixed it; thanks!

PR #1907, opened 2022-11-22
new opengraph card image and favicons

gordoncc:
making favicon consistent with developers and operators, better image for social cards.


pages/_app.js
```diff
@@ -59,6 +60,29 @@ function MyApp({ Component, pageProps }) {
...
+        />
+        <link
+          rel="icon"
+          type="image/png"
+          sizes="16x16"
+          href="/images/favicon-16x16.png"
+        />
+        <link rel="manifest" href="/images/site.webmanifest" />
+      </Head>
+
```
matildepark: These are better off in the Meta.js file so as to not have this information in two places.

gordoncc: fixed

matildepark (approved):

PR #1922, opened 2022-12-06
urbitcoin org

gordoncc:
create /organizations/urbitcoin


thomas-kroes:
Doesn't seem like the right place to put the urbitcoin page – it's not an organization in any sense

PR #1947, opened 2023-01-11
Update Getting Around

gordoncc:
Updated copy to reflect new groups UI, and addressed outdated sections.


thomas-kroes:
Good edits, I will merge it for now, but I would like to go over the copy more generally. 
Now it mentions uqbar's groups specifically, which is fine but I want to calibrate all the different organizations fairly and highlight stuff that works very well so we will have to go over it more in depth anyway.

thomas-kroes (approved):

PR #1948, opened 2023-01-12
Updating Dalten / Quartus related ecosystem pages

thelifeandtimes:
updating the Quartus app listings to occur under ~dister-dozzod-dalten and adding a few dalten/quartus related groups/apps.


thomas-kroes:
Sarlev, happy to merge but you use ~dalten at some apps and Quartus at others, is that correct?

PR #1964, opened 2023-01-17
cli-docs: add runtime upgrade instructions

### LINKED ISSUE #1962: Add dock and runtime binary instructions to CLI getting started guide
Currently the instructions for dock and runtime upgrades are in the operators guide where it's not very easy to find them.

Since these two things are common actions and cause a lot of incoming support requests, I think it's useful to have them in the getting started guide.

Most CLI users won't need to know much more than how to install/boot urbit and upgrade their runtime. Other things they might have to know are in the cloud hosting instructions.

zalberico:
Fixes: https://github.com/urbit/urbit.org/issues/1962

This commit clarifies language around the Urbit runtime. I also add instructions about how to upgrade the runtime via next and if the user is not docked, how to dock.

### DIFF
--- a/content/getting-started/cli.md
+++ b/content/getting-started/cli.md
@@ -4,9 +4,10 @@ weight = 2
 description = "Installation instructions for power users."
 +++
 
-If you're a power user, you can run the Urbit virtual machine directly using the
-command line. This can be run on your local machine or a server in the cloud, we
-just cover the general case here.
+If you're a power user, you can run the Urbit runtime (Vere) using the command 
+line. This can be run on your local machine or a server in the cloud, we
+just cover the local case here. The runtime is what interprets the Urbit kernel
+code (Arvo) into commands your specific machine (macOS or Linux) understands.
 
 Note there is a much more comprehensive [cloud hosting
 guide](https://operators.urbit.org/manual/running/hosting) which walks through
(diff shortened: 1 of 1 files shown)


content/getting-started/cli.md
```diff
@@ -244,8 +246,41 @@ Whichever address and port it says there is the one to open in the browser.
...
+> default to port 8080):
+>
+> ```shell
+> sudo apt-get install libcap2-bin
+> sudo setcap 'cap_net_bind_service=+ep' ./urbit
+> ```
+
+If you've been running Urbit for a while (from before runtime version 1.9) 
+and these `.run` commands don't work for you, it probably means you need to dock 
+your pier. You can do this with the following command:
```
jalehman: What is docking your pier? Is this documented anywhere (e.g. by running `./urbit dock` w/o arguments? It's the first question that popped out to me. Probably wouldn't hurt to have something somewhere that explains what this is doing.

jalehman (commented):
This looks great, tyfs.

content/getting-started/cli.md
zalberico: https://operators.urbit.org/manual/running/vere#dock - I can link the word dock to this.

jalehman: Perfect, exactly what I was looking for.

jalehman (approved):

PR #1971, opened 2023-01-18
Fix setcap command for docked binaries

### LINKED ISSUE #1970: For docked binaries setcap needs to be run on .run not ./urbit
Reported by @rabsef-bicrym - if you have docked your binaries following the instructions in the getting started guide and you're on linux, the setcap command should be run against <pier>/.run not ./urbit

zalberico:
Fixes: https://github.com/urbit/urbit.org/issues/1970

If the binary is docked then setcap needs to run against .run


zalberico:
@jalehman can you approve and merge when you see this?

jalehman (approved):

PR #1974, opened 2023-01-18
Fix setcap command everywhere

### LINKED ISSUE #1970: For docked binaries setcap needs to be run on .run not ./urbit

zalberico:
Fixes: https://github.com/urbit/urbit.org/issues/1970

The previous commit only made this fix in the upgrade section.

This commit moves the setcap instructions to the proper place and
changes the command from being run on the binary to .run.


belisarius222 (approved):
lgtm

PR #1977, opened 2023-01-18
"Focusing the Runtime" windows and port deprecation blog post

gordoncc:


thomas-kroes:
I think we should mention groundseg here, although it hasn't been released by native yet, it will be out in the coming 2 weeks including windows support.

PR #1979, opened 2023-01-19
privacy and tos updates

gordoncc:
Replaced Tlon with Urbit Foundation / the UF, except for in the arbitration clause in the TOS where there's an address for Tlon Corporation, which I think is no longer current.  Do we have an address that's relevant to this clause?

Screenshot 2023-01-19 at 11 13 04 AM


thomas-kroes:
We could use our cayman address? Do we get notified if mail is received for us there? @jalehman

thomas-kroes:
I have just looked at this again and besides the address there is a bunch of US legal information in there.

gordoncc:
Do we need these pages at all since the UF is not a for-profit US company?  I mean, Tlon doesn't have any Privacy or ToS pages and they actually are offering a service.

thelifeandtimes:
Closing PR as stale.

PR #1999, opened 2023-01-30
February Spotlight

gordoncc:
Includes Holium, NativePlanet, and Urbitex.

Also added twitter handles to orgs.


thomas-kroes:
what's going on with vercel?

gordoncc:
> what's going on with vercel?

All sorted

jalehman (changes requested):
Feels a little weird to have this just be three organizations. Tlon especially, given that the only thing really new with them (yet) is Groups 2 and Talk, which was featured in the last spotlight. 

Instead of Holium, I'd rather see Realm featured. Also, the youtube preview is broken. 

For all of these, given that they're a spotlight, I'd expect us to have a lot more to say than two sentences and a picture.

gordoncc:
> Feels a little weird to have this just be three organizations. Tlon especially, given that the only thing really new with them (yet) is Groups 2 and Talk, which was featured in the last spotlight.
> 
> Instead of Holium, I'd rather see Realm featured. Also, the youtube preview is broken.
> 
> For all of these, given that they're a spotlight, I'd expect us to have a lot more to say than two sentences and a picture.

Replaced Holium with Realm and Tlon with a new pod with Trent and added some more copy.

With regard to what we have to say – I've been following the maxim of "show, don't tell", avoiding injecting an editorial perspective, which would demand personal thoughts and opinions on these items.  That could be fun, but it feels like that should be a separate product.

jalehman (changes requested):
The copy and images you added is good, that's about as much as we need.

Featuring Realm and a podcast about Realm is a little much — perhaps mention the podcast in the Realm spotlight section, and pick something different for the third slot? There are a few new apps that dropped as a result of the encode.club hackathon (which you could also mention), and Galen appeared on the Boys Club podcast as well: https://podcasts.apple.com/us/podcast/urbit-for-idiots-galen-wolfe-pauly-tlon-corporation/id1616327892?i=1000597057523

gordoncc:
Fair point on the Trent redundancy.  Added in Boys Club instead.

thomas-kroes (approved):

thomas-kroes (approved):

jalehman (approved):

PR #2001, opened 2023-01-31
Added bounty for generator doccords

vvisigoth:
I'd like feedback from @jalehman on compensation but otherwise gtg.


vvisigoth:
Morgen as admin because got stamp out of band

PR #2003, opened 2023-02-03
Update grants.js

thomas-kroes:


thomas-kroes:
@gordoncc I tried to play around and add the application deadlines in a suitable place. However as you see without success. Could you take a look at where to put the dates best and maybe the grants@urbit.org mail for questions?

gordoncc:
@thomasurbit Do you really want to update the application deadlines every month?  It seems sufficient to say "bi-weekly"

thomas-kroes:
I think it would be good for people to know when they can expect an answer. I'm happy to update that.

gordoncc:
I'd manage expectations by saying you'll hear back within two to three weeks rather than giving specific operational details.

gordoncc (approved):

Issue #1997, opened 2023-01-30
ecosystem: add image/text content to groups app


jalehman:
The Groups ecosystem page is really barebones right now. Would probably make sense to spruce it up a bit.

Issue #1996, opened 2023-01-30
ecosystem: add talk


jalehman:
Talk should have its own page, and should probably come with images.

gordoncc:
Added basic description and image in https://github.com/urbit/urbit.org/pull/1986

PR #2008, opened 2023-02-09
gc/groupstalk

gordoncc:
updating copy and adding videos to groups and talk.

addresses https://github.com/urbit/urbit.org/issues/1997

and https://github.com/urbit/urbit.org/issues/1996


thomas-kroes:
awesome!

thomas-kroes (approved):

PR #2012, opened 2023-02-10
Pony ecosystem page

hanfel-dovned:


thomas-kroes:
Hey, please add a description so it displays better. Also it's not compatible with the current Kelvin version.
`description = ""`

hanfel-dovned:
Current Kelvin version is 415, right? It works on ~hanfel-dovned, which is on tin8p.

jalehman (approved):
Works just fine for me.

PR #2013, opened 2023-02-10
add deks generator grant

vvisigoth:


jfranklin9000:
`+vat` and `+vats` (`+dek` and `+deks`) could be combined into a single generator that takes 0 (`+vats`), 1 (`+vat`), or more desks (like the modification to `|suspend`).
```
+deks
+deks %base
+deks %groups %talk
```

jalehman (approved):
This looks good. The one other thing is that `+vats` prints out so much information that it's unwieldy to work with. That's probably because it was designed before people regularly had very many desks at all. 

A potential addition would be to have `+deks` print less detail by default — e.g. names of all desks, last digits of hash, status, all on one line per desk — with an argument to turn on additional detail. `+dek` would have full detail.

jalehman:
> +vat and +vats (+dek and +deks) could be combined into a single generator that takes 0 (+vats), 1 (+vat), or more desks (like the modification to |suspend).

This is a good idea.

Also, why the name `deks`? Any reason to not use `desk`?

vvisigoth:
Closing in favor of this https://github.com/urbit/urbit/issues/6297

PR #2027, opened 2023-02-20
fix broken urbit.org/grants link

ripnyt-ripnyt:
the current link is relative, and resolves to `https://urbit.org/blog/urbit.org/grants` and returns a 404. I added the protocol to make it consistent with the other links in the blog and expect a corrected behavior.


thomas-kroes:
Thanks!

thomas-kroes (approved):

PR #2032, opened 2023-02-23
grant: remove pals tags integration from whom

Fang-:
I've been overseeing @JohnRillos' work on this, and it's been great. Whom is as good as complete, and checks all the boxes. ...Except the integration with pals' tags metadata.

We probably still want this in at some point, but right now it's blocked on me improving the pals api to provide better access to tags and... I just haven't done that yet. It seems unfair to keep this grant blocked on that, so after brief discussion, we've decided to remove it from the formal spec of the grant.


vvisigoth (approved):
j'approve

PR #2042, opened 2023-03-02
Create RFP grant type

vvisigoth:
This PR creates the RFP grant type. Concretely, this means that there is now an overview page with copy explaining RFPs as well as support for displaying previews of RFPs (reusing the styling from apprenticeships).

Note that this keeps all the existing pages related to apprenticeships, though they will no longer be linked to from /grants.

I'm happy to change any of this copy if it's unclear. I'll open a separate PR with a few initial RFPs.


thomas-kroes:
Few things
- the hackerhouse grants office hours should probably be specified with a time and day, also we need to put it on the other grants pages that the office hours exist. 
- the beginning of the jump in section is the same as the end of the previous section, the latter can probably be deleted
- it doesn't state that the proposal needs to be reviewed in the grant cycle by the team
- I'm wondering whether the proposal form is the right process, probably it's the right way to go but it might make sense to ask Abby about what's the easiest (also for getting metrics)

thomas-kroes:
We need to find a way to filter for proposals, it still says apprenticeships there

thomas-kroes:
the hacker house session starts in the dark room, where often normal meetings are held, as this link will have rando's show up at any given time, it's probably better to select another room for grants/School office hours. office hour room or smt

vvisigoth:
@thomasurbit is going to put the office hours information in the UF group and then we'll edit this with instructions to join the group and where to find the hours and a link.

@thomasurbit the "Requests" filter doesn't show up here because there are no grants with that tag in this PR. I've tested with the content [here](https://github.com/urbit/urbit.org/pull/2043) and the requests tag shows up and filters correctly. The apprenticeships tag will continue to show because there has been previous content tagged with Apprenticeship.

thomas-kroes:
Styling, the test cases you put in are grey and have no coloured tag. We probably should choose (or just merge them with what formerly was apprenticeships)

thomas-kroes:
Just see your previous comment, yes then we might just need a new colour for requests @gordoncc what colour would fit the grant scheme for requests? (tag + card colour)

gordoncc:
@thomasurbit we can use purple like we do on the roadmap for the tag.  bg-purple for the tag, bg-purple-200 for the card background.

vvisigoth:
@thomasurbit @gordoncc there doesn't seem to be a `bg-purple-200`. Where are these styles so I can update.

vvisigoth:
Updates:

- Apprenticeships have been readded with original green styling
- Requests now use purple style
- I updated package.json to use the most recent design system to get purple style. FYI to run `npm i` to update this locally.

Looks like this with the dummy request (remember that the RFP content is in another PR)

image

Please review @jalehman @thomasurbit

thomas-kroes:
This looks great, as you might see I tried to make some small changes which resulted in this mess. It's all fixed by now and I'm on mandatory git training.

vvisigoth:
@thomasurbit so do you approve? If so, please approve formally using button at top right.

PR #2054, opened 2023-03-21
Show in-progress grants on /grants

vvisigoth:
APropos of a discussion with @jalehman we want to also include In Progress grants on the default grants view to make sure people can see what is in flight.


thomas-kroes:
Any way we can sort on creation time + open/in progress? Would be good to display the open grants on top and the in progress ones below.

vvisigoth:
@thomasurbit I'm ambivalent on sort order. @jalehman what say you?

PR #2065, opened 2023-03-31
april updates

thomas-kroes:


alangalis (approved):
Looks good to me. (We've got a pretty reliable schedule going lately. Could be easier to update this in the future to say we review on every 2nd and 4th wednesday so we don't need to continue to update.)

thomas-kroes:
Sounds like a plan to me!

PR #2082, opened 2023-04-07
updating featured card with new HSL link

gordoncc:
I updated developers.urbit.org to include a link to a featured c2a to the next hoon school since the current one was out of date.  Doing the same on urbit.org.


thomas-kroes (approved):
Thanks good catch

PR #2098, opened 2023-04-20
update to history

thomas-kroes:


gordoncc (approved):

tacryt-socryp:
Recommendation to do the year timeline in reverse chronological order bc the early history is far less important to most audiences and can be distracting/present an incorrect impression (y'all have been working on this for how long???) @thomasurbit @jalehman

thomas-kroes:
@tacryt-socryp I agree with you on this one.

PR #2102, opened 2023-04-25
Update 2023-04-27-urbit-meetup-berlin.md

gordoncc:
Fix formatting to not use dark background


thomas-kroes:
image just didn't load, should work better like this!

thomas-kroes (approved):

PR #2121, opened 2023-05-11
%loch proposal changed to %lick

mopfel-winrux:
I Modified the proposal to represent the work beign done with %lick


vvisigoth (approved):
LGTM

PR #2130, opened 2023-05-22
add event amsterdam

thomas-kroes:


thomas-kroes:
was a test for a video

PR #2144, opened 2023-06-02
Fix Landscape link on interfaces page

dnbrwstr:
An update to Tlon's website removed the standalone landscape page, breaking the link here. 

This PR fixes that issue by linking to Tlon's homepage instead.


thomas-kroes (approved):

thomas-kroes:
Thanks for flagging!

PR #2163, opened 2023-06-21
Replit bounty

sigilante:


jalehman (approved):
Who will be championing this? Seeing as it's a bounty, I'd expect your or someone else from our team. Otherwise this LGTM.

sigilante:
Thomas, Matt, and I decided on Matt since I'll be out in July and we want this moving before then.

PR #2198, opened 2023-08-09
Urbit Argentina Meetup #2

mnrrxyz:
File for the 2nd Meetup in Argentina.


thomas-kroes:
we were too late in catching this one, please add me as a reviewer next time so I get notified and process it on time or send me a dm.

mnrrxyz:
No prob, will do. Can it still be added in the "Past Events" section as proof that it happened?

PR #2226, opened 2023-10-12
Create h00ncr3w.md

narskidan:
Per the instructions at https://urbit.org/groups/submit, I'm attempting to add my group to the directory.


narskidan:
I've also now sent the requisite DM to ~haddef-sigwen

Screenshot 2023-10-12 at 0 03 13

narskidan:
Huh, it seems ~haddef-sigwen no longer has access to push to this repo. Is the Groups directory frozen, or is someone else in charge of it now (in which case, the web page should be updated to say so)?

PR #2253, opened 2024-03-06
Fix broken urbit.org install links

pkova:


Fang- (approved):
These look like the appropriate redirects, yeah.

PR #2258, opened 2024-03-07
twitter cards change

thomas-kroes:


ripperi (approved):

ripperi:
Since we prefer the dark theme, should we inverted the colors here? black on white -> white on black

PR #2267, opened 2024-03-27
Git on Urbit: supplemental milestone A -- server push support

mikolajpp:
A proposal to complete Urbit git hosting capability by implementing support for pushing to an Urbit-hosted git repository.


tamlut-modnys (approved):
Neal says this change can be approved.

PR #2321, opened 2024-07-02
Malformed newlines in chat bots grant

narskidan:
Currently, Milestone 2 has a section that looks like this:

Screenshot 2024-07-02 at 2 40 45 a m

Markdown mungs single newlines unless you add two spaces at the end. This PR adds the spaces so that the newlines render correctly.


narskidan:
How it looks with these changes applied:

Screenshot 2024-07-02 at 2 42 17 a m

tamlut-modnys (approved):
Looks good

PR #2334, opened 2024-09-02
Create Subassembly PNW listing

thelifeandtimes:
Adds an entry to urbit.org for the Subassembly event.

I also added a property, `event_details`. Would be preferred to add a button to event pages that use this property (next to the 'register' button.


thelifeandtimes:
Realized `links` property actually afforded the button I was looking for from `event_details`.

tamlut-modnys:
MERGED

tamlut-modnys (approved):
GOOD

PR #2354, opened 2024-10-18
Adding New England Meetup event 5 to urbit.org

7Sages7:
Adding New England Meetup event 5 to Urbit.org as part of my grant with UF


tamlut-modnys (approved):
Looks good to me.

PR #2368, opened 2025-11-11
Replace codebase with new site code from urbit/urbit.org-reloaded

thelifeandtimes:
This replaces the urbit.org repo with the new site version that was initially developed in a dedicated repository. The previous version is available in the backup-before-reloaded branch.


kennyrowe (approved):
New website

PR #2464, opened 2026-08-25
Fix grammar in contributor spotlight

bitemyapp:
Fixes the malformed sentence in the ~navsul-pagrec contributor spotlight so it reads:

> but that gives us strong evidence over the corpus, not a proof covering all possible Hoon.

Updates both checked-in copies of the article.

Targets the head branch of #2463.

Verification: `git diff --check`


thelifeandtimes:
directly incorporated changes into #2463

